Issue #212, opened 2026-04-24
Build Studio: implement the Typecheck Gate (per-task and pre-ship)


markdbodman:
## Context

AGENTS.md §"Typecheck Gate — Build Studio (policy + implementation status)" mandates that Build Studio apply the same typecheck + production-build gate to the sandbox it builds in as the local pre-commit hook applies to human commits. **Neither half is currently implemented** (audited 2026-04-24).

Today, CI still catches typecheck/build failures when a BS-produced PR opens — the `main` branch protection gates hold — but the goal is to catch them at the sandbox boundary so a red build never escapes Build Studio.

### 1. Per-task verification

**Current state:** [apps/web/lib/integrate/build-orchestrator.ts:375-401](../blob/main/apps/web/lib/integrate/build-orchestrator.ts#L375-L401) classifies task outcome by string-matching the CLI's stdout for keywords like `"error"` / `"typecheck.*fail"`. A task whose CLI summary omits those words can silently pass even if typecheck is red.

**Fix:** After `classifyOutcome()` returns at [build-orchestrator.ts:583](../blob/main/apps/web/lib/integrate/build-orchestrator.ts#L583) and before persisting status `"DONE"`, exec two commands inside the sandbox container:

```
pnpm --filter web typecheck
pnpm --filter web build
```

On non-zero exit from either, override the task to `"BLOCKED"` with the command output appended to the coworker context so the next iteration can read and fix it.

### 2. Pre-ship verification

**Current state:** [apps/web/lib/queue/functions/build-review-verification.ts](../blob/main/apps/web/lib/queue/functions/build-review-verification.ts) runs browser-use UX tests only. No typecheck or build.

**Fix:** Add two sibling `step.run(...)` blocks that exec the same two sandbox commands, persist their pass/fail + stdout tails alongside `uxTestResults` on `FeatureBuild`, and extend `checkPhaseGate` (called from `advanceBuildPhase` in `apps/web/lib/actions/build.ts`) to block ship until both are green.

## Acceptance criteria

- [ ] A BS task whose diff introduces a TypeScript error ends `"BLOCKED"` with the `tsc` output in the context, not `"DONE"`.
- [ ] A BS task whose diff fails `next build` (e.g. missing import, broken route) ends `"BLOCKED"` with the build output in the context.
- [ ] A build in `review` phase with a typecheck-red diff cannot advance to `ship`; the ReviewPanel surfaces the error clearly.
[...]

PR #213, opened 2026-04-24
docs(agents): flag BS Typecheck Gate as unimplemented vs. mandated

markdbodman:
Audit on 2026-04-24 confirmed neither half of the "Typecheck Gate — Build Studio" section is actually implemented:

- Per-task verification does not run typecheck/build; build-orchestrator classifyOutcome() just string-matches the CLI stdout for "error" / "typecheck.*fail". A task can silently pass even with a red typecheck if the CLI summary omits those keywords.
- Pre-ship verification in build-review-verification.ts runs browser-use UX tests only. No typecheck or build step.

Rather than weaken the policy to "planned", keeps the full mandate and adds an explicit "Implementation status" block that:
- Names what's missing, file:line accurate
- Names the landing spots for the real fix
- States the interim risk (sandbox can hand a typecheck-red build to CI; CI still catches it, but sandbox gate is the stated goal)

Preserves governance intent while being honest about reality. A separate GitHub issue will track the actual implementation work.

## Changes

-
-

## Test plan

- [ ] `pnpm typecheck`
- [ ] `pnpm test`
- [ ] `pnpm --filter web build`
- [ ] Manual verification (describe below)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -116,14 +116,21 @@ New clones inherit this automatically via the repo's `postinstall` script. Verif
 
 **Subagent dispatchers:** include "run `pnpm --filter web typecheck` before committing and fix any errors" in every task prompt that modifies TypeScript. Subagents do not read AGENTS.md.
 
-### Typecheck Gate — Build Studio (mandatory)
+### Typecheck Gate — Build Studio (policy + implementation status)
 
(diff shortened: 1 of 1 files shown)


markdbodman:
Tracking issue for the actual implementation work: #212

Issue #219, opened 2026-04-24
Unit Tests: 9 renderToStaticMarkup failures on CI ('Invalid hook call' / 'Cannot read properties of null (reading useState)')


markdbodman:
## Summary

Nine tests reliably fail in CI with `TypeError: Cannot read properties of null (reading 'useState')` but pass locally. All are components that use React hooks rendered via `renderToStaticMarkup` from `react-dom/server`.

Seen on [run 24876320210 for #217 merge to main](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/24876320210/job/72833657582) and every PR CI run since #203 merged on 2026-04-24.

## Affected tests

- `apps/web/components/admin/ForkSetupPanel.test.tsx` (7 cases)
- `apps/web/components/platform/ToolExecutionLogClient.test.tsx` (1 case — `shows the GAID identity reference when present`)
- `apps/web/app/(shell)/platform/identity/page.test.tsx` (1 case — `renders the identity workspace landing page with core management cards`)

## Stack trace (CI)

```
TypeError: Cannot read properties of null (reading 'useState')
  at process.env.NODE_ENV.exports.useState ../../node_modules/react/cjs/react.development.js:1263:33
  at ForkSetupPanel components/admin/ForkSetupPanel.tsx:37:35
  at Object.react_stack_bottom_frame ../../node_modules/react-dom/cjs/react-dom-server-legacy.node.development.js:9808:18
```

Line 1263 is `useState = resolveDispatcher().useState(...)`. Dispatcher is `null` because `ReactSharedInternals.H` is unset when the component's React copy differs from the copy that `react-dom/server` called into.

## Root cause (suspected)

Multiple React copies in the module graph. `apps/web/vitest.config.ts` already has explicit path aliases and `resolve.dedupe: ["react", "react-dom"]`, but `react-dom-server-legacy.node.development.js` does a CJS `require("react")` that may resolve to a different copy than the ESM-imported `react` in the component. Hoisting layout differs between local Windows and CI Linux pnpm installs, which is why local passes and CI fails.

## Why this isn't test drift

Unlike #216/#217/#218, the source strings still match the assertions — the components just fail to render at all. The right fix is test-infra, not component-text.

## Options worth trying

- Add `@testing-library/react` + `jsdom` as dev deps and convert these tests to `render()` with `environment: "jsdom"`. Matches how the rest of the ecosystem tests client components.
[...]

PR #223, opened 2026-04-24
fix(test): exact-match react SSR aliases in Vitest

markdbodman:
## Summary
- switch React and React DOM Vitest aliases to exact-match regex entries
- keep the existing SSR inline handling while avoiding prefix alias rewrites of react-dom subpaths
- address the Linux CI unit-test failure where react-dom/server rendered against a different React instance

## Verification
- pnpm test
- pnpm typecheck
- cd apps/web && npx next build

### DIFF
--- a/apps/web/app/(shell)/platform/identity/page.test.tsx
+++ b/apps/web/app/(shell)/platform/identity/page.test.tsx
@@ -1,5 +1,8 @@
+// @vitest-environment jsdom
+
+import { act } from "react";
+import { createRoot } from "react-dom/client";
 import { describe, expect, it, vi } from "vitest";
-import { renderToStaticMarkup } from "react-dom/server";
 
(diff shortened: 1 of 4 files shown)


markdbodman:
CI still fails with the same 9 `Cannot read properties of null (reading 'useState')` errors on [run 24877375245](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/24877375245/job/72837018832). Combined with #221 (which tried `server.deps.inline`) also failing, the root cause isn't the alias shape or the SSR transform pipeline — something else is handing react-dom-server a different React copy than the component imports on CI's Linux pnpm layout.

Escalating back to the other options in #219: converting to `@testing-library/react` + jsdom, or explicitly aliasing `react-dom/cjs/*` subpaths. Closing-and-retrying isn't useful if we're still guessing at the config-level fix.

PR #519, opened 2026-05-12
fix(mcp): harden connectivity onboarding

markdbodman:
## Summary
- add explicit `type: "http"` to generated Claude Code, Codex, and VS Code MCP setup snippets
- harden the MCP JSON-RPC transport gate so plain-HTTP host authorization ignores caller-controlled `X-Forwarded-Host`
- keep `X-Forwarded-Proto: https` support for TLS-terminated proxy paths and retain `MCP_INSECURE_INTERNAL_HOSTS` for explicit internal Docker hosts

## Verification
- `pnpm --filter web exec vitest run app/api/mcp/v1/route.test.ts lib/actions/mcp-tokens.test.ts`
- `pnpm --filter web typecheck`
- `pnpm --filter web build` (passes; existing Turbopack/NFT broad-trace warnings remain)
- `git diff --check`

## Notes
- This is the first implementation slice after the reviewed agent workspace pattern docs: MCP connectivity/onboarding repair only.
- A model-mediated sandbox CLI probe still needs to be rerun after this branch is deployed into the running portal image; this PR covers the generated config and route guard that blocked that path.

### DIFF
--- a/apps/web/app/api/mcp/v1/route.test.ts
+++ b/apps/web/app/api/mcp/v1/route.test.ts
@@ -96,11 +96,7 @@ describe("POST — transport guards", () => {
     expect(res.status).toBe(401); // got past TLS guard, failed at auth
   });
 
-  // Regression: when Next.js runs inside a container or behind a proxy,
-  // request.url reflects the *internal* bind address (e.g. 0.0.0.0). The
-  // transport guard must consult the forwarded host/proto headers — what
-  // the client actually connected to — not the URL Next reconstructed.
-  it("allows containerized request when X-Forwarded-Host points to localhost", async () => {
+  it("rejects spoofed X-Forwarded-Host on plain HTTP", async () => {
     resolveMock.mockResolvedValue(null);
     const res = await POST(
       makeRequest({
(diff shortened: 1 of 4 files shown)


markdbodman:
Additional production-path verification after opening the PR:

- Rebuilt/recreated `dpf-portal-1` from this branch with `docker compose --env-file D:\DPF\.env -p dpf -f D:\DPF\.worktrees\mcp-connectivity-onboarding\docker-compose.yml up -d --no-deps --build --force-recreate portal`
- Confirmed rebuilt portal is healthy and labeled with this branch worktree compose file
- Confirmed `MCP_INSECURE_INTERNAL_HOSTS=portal,host.docker.internal,sandbox` is present in the running portal container
- Host spoof probe: `Host: evil.example.com` + `X-Forwarded-Host: localhost:3000` returns `403`, so X-Forwarded-Host no longer bypasses the plain-HTTP guard
- TLS proxy probe: `X-Forwarded-Proto: https` reaches auth and returns `401` for an invalid token, preserving the TLS-terminated proxy path
- Internal bridge probe from `dpf-sandbox-1` to `http://portal:3000/api/mcp/v1` reaches auth and returns `401` for an invalid token, preserving internal Docker HTTP access
- Authenticated JSON-RPC probe using the existing local MCP token returned `HTTP 200`, `jsonrpc: 2.0`, `34` tools, including `list_epics`
- Rebuilt image source contains all three generated MCP setup snippets with `type: "http"`

PR #793, opened 2026-05-19
fix(deliberation): return null from parseReviewResponse on parse failure

markdbodman:
## Summary

- `parseReviewResponse` now returns `null` when the LLM response can't be parsed (empty, malformed JSON, rate-limit garbage) instead of a synthetic `{decision:"fail", issues:[critical:"unparseable"]}` result
- Both `mcp-tools.ts` dual-reviewer blocks push parse-failed reviewers as `{review:null, failureReason:"unparseable"}` branches — the synthesizer marks these `completed:false` and excludes them from the consensus vote
- Single-reviewer callers in `build.ts` null-coalesce to the same fail fallback they had before

## Root cause

Observed 2026-05-19 during FB-7A21E1F6 ideate phase: Codex weekly token limit caused Reviewer 2 to return garbage JSON. `parseReviewResponse` returned `{decision:"fail"}`, which was counted as a legitimate fail vote. With Reviewer 1 saying "pass" and Reviewer 2 saying "fail", the deliberation correctly computed `consensusState:"no-consensus"` — but this was a false disagreement, not a genuine design conflict. The audit trail was polluted and the consensus state was misleading.

## Test plan
- [ ] `build-reviewers.test.ts` — 23 tests pass (verified locally)
- [ ] `parseReviewResponse("not JSON")` now returns `null`
- [ ] Valid pass/fail responses still return `ReviewResult` as before
- [ ] Dual-reviewer deliberation with one parse failure → `partial-consensus` (one reviewer) rather than `no-consensus` (fake disagreement)


markdbodman:
**Re: CodeQL alerts 40 & 41 (js/tainted-format-string)**

Both alerts are false positives on pre-existing code.

**Why they appeared:** My changes to `build.ts` triggered CRLF→LF line-ending normalization across the whole file, making git re-diff every line. The flagged `console.error` calls at lines 573 and 677 are identical to what's on `main` — only the line endings changed.

**Why they're false positives:** JavaScript template literals are not format strings. `console.error(`...${buildId}...`, err)` performs string interpolation; Node.js does not interpret `%s`/`%n` specifiers. Even if `buildId` were attacker-controlled, it goes only to server logs visible to admins. The value is an internal DB ID (format `FB-XXXXXXXX`), not user-provided.

Dismissed both alerts as *false positive* in the Security tab.

markdbodman:
**Re: Unit Tests failure**

The two failing tests (`mcp-tools.thread.test.ts` and `AiOperationsMap.test.tsx`) are **pre-existing failures on `main`** — all three of the last CI runs on main show Unit Tests: failure (runs [26078948985](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/26078948985), [26079226490](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/26079226490), [26080060688](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/26080060688)). Neither file is touched by this PR.

**Typecheck** fix just pushed — I had added `!` assertions to `build-reviewers.test.ts` but missed the same pattern in `build-disciplines-integration.test.ts`. Both are now consistent.

markdbodman:
Closing as superseded by #794, which fixed the same bug with a cleaner approach (typed `parseError?: true` flag on ReviewResult rather than nullable return). The two PRs were developed concurrently and both landed on the same root cause. No new value to add here — #794 is already merged and CI on main now reflects the correct behaviour.

PR #796, opened 2026-05-19
feat: guard Build Studio sandbox source currency

markdbodman:
## Summary
- add Build Studio sandbox source-currency classification before build branch work starts or resumes
- auto-refresh clean stale sandbox branches from the target ref and pause dirty, diverged, or unknown states with visible BuildActivity/sourceCurrency evidence
- preserve sourceCurrency in build execution state and surface it through the sandbox projection and BuildSandboxCard UI
- keep the source-copy bootstrap from overwriting an already-prepared git workspace
- make Build Studio progress loading tolerate partial flow-state fetch failures

## Why
The Build Studio sandbox was able to keep developing against stale main without a visible source-currency contract. That made it hard to know whether a build branch was current, behind, diverged, or silently stale before dispatching new coworker work.

## Verification
- pnpm --filter web exec vitest run lib/integrate/sandbox/sandbox-source-strategy.test.ts lib/integrate/sandbox/sandbox-source-currency.test.ts lib/integrate/sandbox/build-branch.test.ts lib/build/sandbox-state.test.ts lib/integrate/build-pipeline.test.ts lib/actions/build-governed.test.ts lib/build/progress-visibility.test.ts lib/mcp-tools-build-observer.test.ts components/build/BuildSandboxCard.test.tsx components/build/BuildProgressOperationalPanel.test.tsx
  - 10 test files passed, 82 tests passed
- pnpm --filter web typecheck
  - passed
- pnpm --filter web build
  - passed; existing broad Turbopack trace warnings remain

## Portal validation
Per operator direction, production-path portal validation will happen after this PR is processed and the portal is updated to this branch.

### DIFF
--- a/apps/web/lib/actions/agent-coworker.ts
+++ b/apps/web/lib/actions/agent-coworker.ts
@@ -715,12 +715,32 @@ export async function sendMessage(input: {
     // Inject Build Studio context — use explicit buildId or auto-resolve on /build route.
     // Exclude `abandoned` so a prior-test abandoned build doesn't shadow the real active build.
     if (!resolvedBuildId && input.routeContext.startsWith("/build")) {
-      const latestBuild = await prisma.featureBuild.findFirst({
-        where: { createdById: user.id!, phase: { notIn: ["complete", "failed", "abandoned"] } },
-        orderBy: { updatedAt: "desc" },
-        select: { buildId: true },
-      });
-      resolvedBuildId = latestBuild?.buildId ?? undefined;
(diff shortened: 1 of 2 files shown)


markdbodman:
Closed as a stale duplicate of #784. The old head would have removed newer runtime-target registration now present on main, so the branch was rebased by skipping the already-applied source-currency commit, force-updated to current main with 0 changed files, and then the stale branch/worktree were cleaned up.

PR #997, opened 2026-05-22
feat(mcp): role-based token templates (Admin / Development / Employee / Observer)

markdbodman:
## Summary

Refs [BI-9866659C](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory) — MCP token lifecycle UX. This PR lands AC #2 (named templates) and the renaming concern; idle hygiene, rotate-with-edit, ephemeral ship-phase tokens, and revoked-token hide are deferred to follow-up.

Admin > Platform Development > MCP previously asked operators to compose a token by ticking individual grants against the coarse read/write/admin tier. Operators have to think about ~40 grant strings to issue one token, and the prominent **"Issue write token"** button hides what is functionally a development-agent bundle.

This PR adds a role template catalog so issuance becomes \"pick a role\" instead of \"pick 30 grants correctly\".

### Templates (all grants drawn from the existing TOOL_TO_GRANTS catalog — no new grant strings)

| Template | Tier | Bundle |
|---|---|---|
| **Admin** | admin | every grant including admin_write, policy_write, iac_execute |
| **Development** | write | code/specs/backlog reads, work-capsule writes, sandbox_execute, iac_execute, build_promote, release gates, deliberation |
| **Employee — Finance** | write | financial_report_create + reads |
| **Employee — HR** | write | consumer_read/write, policy_write, reads |
| **Employee — Enterprise Architecture** | write | ea_graph_read/write, document_publish |
| **Employee — Marketing** | write | marketing_read/write, reads |
| **Observer** | read | every *_read grant, no writes anywhere |
| **Custom** | user-picked | flat checkbox picker (legacy UX preserved) |

Each template's grants are filtered against the install's registered scope catalog at issue time, so installs that don't expose a given grant won't be issued a token requesting it.

### Renaming

The misleading **\"Issue write token\"** affordance becomes **\"Issue development token\"** since the bundle is the coding-agent loadout, not a generic write scope. \`issueMyWriteMcpToken\` is preserved as a back-compat alias that delegates through the development template — no callers break.

### UX changes

- Primary button: **\"Issue development token\"** (one-click 90-day dev bundle)
- Secondary button: **\"Issue token from template\"** opens the role picker modal
- Modal puts the role template dropdown above name/expiry; grants render as read-only chips when a template is active, flat checkbox picker only when Custom is selected
- Manually toggling a grant flips the picker to Custom so the label stays honest

## Test plan

- [x] \`pnpm --filter web exec vitest run\` for the three affected files: **45 tests pass**
[...]

PR #1012, opened 2026-05-22
feat(mcp): token idle hygiene — last-used / idle-days, stale filter, bulk revoke

markdbodman:
## Summary

Continues [BI-9866659C](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory) — MCP token lifecycle UX. This PR lands **AC #1**: idle-days columns + stale filter + bulk-revoke. AC #2 (named templates) shipped in #997. Remaining ACs (rotate-with-edit, ephemeral ship-phase tokens, revoked-token hide/archive) follow in separate focused PRs.

## What changes

**Server (\`apps/web/lib/actions/mcp-tokens.ts\`, \`apps/web/lib/mcp-token-scopes.ts\`)**
- \`listMyMcpTokens\` now derives \`idleDays\` per row from \`lastUsedAt\` vs \`now\`. Null for never-used / revoked / expired so the UI can render \"—\" instead of an ever-growing number on dead rows.
- New \`bulkRevokeMyMcpTokens(tokenIds, reason)\` action: single ownership lookup for the whole batch, dedups ids defensively, treats already-revoked as idempotent success, surfaces per-token \`status: \"revoked\" | \"already_revoked\" | \"not_found_or_not_yours\" | \"error\"\` so the UI never silently drops a request.
- \`deriveIdleDays\` + \`MCP_TOKEN_DEFAULT_STALE_DAYS = 7\` live in \`mcp-token-scopes.ts\` rather than the actions file — Next.js rejects non-async exports from \`\"use server\"\` modules (caught by the build gate on the first iteration of this PR).

**UI (\`apps/web/components/admin/McpTokenManager.tsx\`)**
- **Stale pill** on the token-name row when \`idleDays ≥ 7\` (or \`createdAt ≥ 7d\` ago for never-used rows still hanging around). Shows the idle duration.
- **Inline annotation** next to Last used: \`(today)\` / \`(Nd idle)\`.
- **Toolbar** above the list with the **\"Show stale only\"** toggle (default off) and a \"stale X of Y\" count.
- **Per-row checkbox** on active rows only (revoked / expired rows aren't selectable).
- **\"Revoke N selected\"** button appears once ≥1 row is selected; confirms before firing; clears selection on success and shows a notice with the revoked/failed counts.

**Why 7 days?** Per the kernel principle [idle-is-not-abandoned](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/blob/main/docs/founder-kernel/wiki/principles/idle-is-not-abandoned.md) — absorbs travel + weekly rate-limit resets so paused work survives a full week without churn.

## Test plan

- [x] \`pnpm --filter web exec vitest run\` for the affected files: **62 tests pass**
  - \`mcp-token-scopes.test.ts\` — template catalog (unchanged from #997) **plus** the existing scope tests still pass with the relocated \`deriveIdleDays\` / \`MCP_TOKEN_DEFAULT_STALE_DAYS\` exports
  - \`mcp-tokens.test.ts\` — \`deriveIdleDays\` (never-used / same-instant / future / fractional / older), \`listMyMcpTokens\` idle-days for active/never-used/revoked/expired with frozen clock, \`bulkRevokeMyMcpTokens\` (auth, empty input, dedup, idempotent already-revoked, ownership rejection, partial failure)
  - \`McpTokenManager.test.tsx\` — inline idle annotation, stale pill rendering only on ≥7d rows, \"Show stale only\" filter narrows the visible list, bulk-revoke wires the right ids through, revoked rows have no checkbox
- [x] \`pnpm --filter web exec next build\` — succeeds (single pre-existing Turbopack NFT cascade warning on next.config.mjs, unrelated)
- [x] Pre-commit typecheck passes
- [ ] **Live UX verification deferred to post-merge** — same pattern as #997. Self-upgrade after merge picks up the new bundle and I'll drive the new toolbar (stale toggle → bulk-select → revoke flow) on \`localhost:3000/admin/platform-development\` then post screenshots back to this thread.

## Build-error caught + fixed mid-PR

First build attempt failed with \"Export not found … The module has no exports at all\" on \`mcp-tokens.ts\`. Root cause: I'd added \`export function deriveIdleDays(...)\` and \`export const MCP_TOKEN_DEFAULT_STALE_DAYS = 7\` to a \`\"use server\"\` file. Next.js requires every export from a server-action module to be an async function — adding non-async exports invalidates the entire file's exports. Moved both to \`mcp-token-scopes.ts\` (a plain module). Worth documenting because the failure mode is generic-looking (\"module has no exports\") even though the cause is specific.

## Out of scope (remaining BI-9866659C)

- One-click \"Rotate with edit\" (atomic issue-new + revoke-old + refresh + show plaintext)
- Per-build ephemeral tokens auto-issued on ship-phase entry
- Show-revoked toggle + 90-day auto-archive


markdbodman:
## Post-merge live UX verification ✅

Rebuilt portal containers from `origin/main` (`c7b19b454`, has all 5 BI-9866659C PRs). Drove the page at `http://localhost:3000/admin/platform-development`.

**Idle hygiene confirmed:**
- Toolbar renders **"Show stale only (≥ 7 days idle — 0 of 23)"** with checkbox + clock icon
- Active rows display inline idle annotation on the Last used line: `(today)`, `(2d idle)`, `(3d idle)` observed on different tokens
- Stale threshold matches `MCP_TOKEN_DEFAULT_STALE_DAYS` (7)
- No tokens currently >= 7d idle on this install, but the filter pipeline is wired through and the counter reads correctly
- Bulk-select checkboxes render on all 23 active rows; suppressed on the 11 revoked rows (verified by toggling Show revoked)

Live page is behaving exactly as the unit tests asserted. Closing the loop on AC #1.

PR #1016, opened 2026-05-22
feat(mcp): rotate-with-edit — atomic issue-new + revoke-old with edited grants

markdbodman:
## Summary

Continues [BI-9866659C](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory) — MCP token lifecycle UX. Lands **AC #3** (rotate-with-edit). #997 shipped templates (#2), #1012 shipped idle hygiene (#1). Remaining: ephemeral ship-phase tokens (#4), revoked-token hide / 90-day archive (#5).

## Why

Today the operator either rotates with the same grants (one-click) or issues a brand-new token from scratch. The common case — \"I need to add \`iac_execute\` to this token without abandoning its name and audit lineage\" — costs a manual issue + manual revoke + manual \`.mcp.json\` copy. Operators reach for \"just mutate the row\" to skip the friction, which is why we keep getting asked for the immutability exception.

Rotate-with-edit closes that gap **without violating immutability**. The McpToken row stays untouched audit-wise — we issue a new token with the edited grants and revoke the old in one atomic operator action.

## What changes

**Server (\`apps/web/lib/actions/mcp-tokens.ts\`)**
- New \`rotateMyMcpTokenWithEdit(tokenId, name, scope, scopes, expiresInDays, baseUrl)\` action.
- Ownership + active-state guard. Refuses revoked or expired rows — those scenarios route through the issue flow instead.
- **Order matters:** issue new token first, then revoke old. Doing it the other way would leave the operator with zero working tokens if issuance failed.
- **Partial-failure handling:** if the new token issues but the old token revoke fails, the new token is still surfaced as live, with \`oldTokenRevokeError\` flagged so the UI can prompt for a manual cleanup revoke. **Never silently drop the new token, never end up with neither.**
- Name falls back to the old token's name when operator leaves it blank, preserving identifiability across rotations.

**UI (\`apps/web/components/admin/McpTokenManager.tsx\`)**
- New **\"Rotate with edit\"** button on each active token row, alongside the existing **\"Rotate token\"** (which still does one-click rotate-with-same).
- Reuses the existing \`TokenFormDialog\`: opens with Custom template, pre-populated with the row's current scope / scopes / name, 90-day expiry default. Operator adds/removes grants like an edit, submits, and sees the new plaintext + setup snippets in the result dialog.
- \`TokenFormDialog\` re-titles to **\"Rotate token with edited grants\"** in rotate mode and the submit CTA reads **\"Rotate with edit\"** / **\"Rotating...\"**.
- \`IssuedTokenDialog\` gains a \`rotated-with-edit\` mode label and renders a **warning banner** with the underlying error when \`oldTokenRevokeError\` is set.

## Test plan

- [x] \`pnpm --filter web exec vitest run\` for the three affected files: **73 tests pass** (+11 for rotate-with-edit)
  - **Action**: auth / empty scopes / not-owned / revoked / expired / happy path with **issue-before-revoke ordering enforced** / name fallback when blank / partial-revoke-failure returns \`oldTokenRevokeError\` / issue-failure short-circuits without touching revoke
  - **Component**: row \"Rotate with edit\" button opens the dialog pre-populated from the row (name + scope + scopes), submit dispatches \`rotateMyMcpTokenWithEdit\` with the right payload, partial-revoke warning banner renders when \`oldTokenRevokeError\` is set
- [x] \`pnpm --filter web exec next build\` — succeeds (only the pre-existing Turbopack NFT cascade warning, unrelated)
- [x] Pre-commit typecheck passes
- [ ] **Live UX verification deferred to post-merge** — same pattern as #997 / #1012. Self-upgrade picks up the new bundle and I drive the new flow on \`localhost:3000/admin/platform-development\` then post screenshots back to this thread.

## Out of scope (remaining BI-9866659C)

- Per-build ephemeral tokens auto-issued on ship-phase entry (#4)
- Show-revoked toggle + 90-day auto-archive (#5)

### DIFF
--- a/apps/web/components/admin/McpTokenManager.test.tsx
+++ b/apps/web/components/admin/McpTokenManager.test.tsx
@@ -1,7 +1,7 @@
 // @vitest-environment jsdom
 
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
-import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { cleanup, fireEvent, render, screen, waitFor, within } from "@testing-library/react";
 
 vi.mock("@/lib/actions/mcp-tokens", () => ({
   bulkRevokeMyMcpTokens: vi.fn(),
@@ -14,6 +14,7 @@ vi.mock("@/lib/actions/mcp-tokens", () => ({
   listMyMcpTokens: vi.fn(),
   revokeMyMcpToken: vi.fn(),
   rotateMyMcpToken: vi.fn(),
+  rotateMyMcpTokenWithEdit: vi.fn(),
   upgradeMyMcpTokenForCodingAgent: vi.fn(),
 }));
 
@@ -25,6 +26,7 @@ import {
   listMcpTokenTemplates,
   listMyMcpTokens,
   rotateMyMcpToken,
+  rotateMyMcpTokenWithEdit,
 } from "@/lib/actions/mcp-tokens";
 import { McpTokenManager } from "./McpTokenManager";
 
@@ -35,6 +37,7 @@ const templatesMock = listMcpTokenTemplates as unknown as ReturnType<typeof vi.f
 const templateIssueMock = issueMyTemplateMcpToken as unknown as ReturnType<typeof vi.fn>;
 const tokensMock = listMyMcpTokens as unknown as ReturnType<typeof vi.fn>;
 const rotateMock = rotateMyMcpToken as unknown as ReturnType<typeof vi.fn>;
+const rotateWithEditMock = rotateMyMcpTokenWithEdit as unknown as ReturnType<typeof vi.fn>;
 
 beforeEach(() => {
   vi.resetAllMocks();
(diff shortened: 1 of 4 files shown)


markdbodman:
## Post-merge live UX verification ✅

Drove **Rotate with edit** on the live install at `http://localhost:3000/admin/platform-development` against the *Finance coworker UX verify* token (issued during #997's verification).

**Confirmed end-to-end:**
- New **Rotate with edit** button renders on every active row, alongside the existing **Rotate token** + **Revoke** buttons
- Clicking opens the modal titled **"Rotate token with edited grants"** with the description: *"Add or remove grants. On submit, a new token is issued with these grants and the original token is revoked — atomically."*
- Name field pre-populated: `Finance coworker UX verify` (from the row)
- Scopes block reads **"Scopes (10 selected)"** with the row's existing grants pre-checked in the flat picker (architecture_read, backlog_read, code_graph_read visible above the fold)
- Role template dropdown set to **Custom** (so the operator sees the flat checkbox picker)
- Submit CTA reads **Rotate with edit** with the refresh icon
- Expiry defaults to **In 90 days**
- Cancel closes cleanly without touching the underlying token

**Minor cosmetic finding** (separate follow-up worth filing): the Role template dropdown labels itself "Custom (read)" even though the row's actual scope tier (and `formScope`) is `write`. The label shows the template's default tier rather than the operator's active tier. The issuance uses `formScope` correctly so the resulting token is the right tier — purely a label cosmetic. Will file a small UX cleanup BI.

End-to-end submit not exercised on this token (no need to actually rotate it). The unit tests already cover the issue-new + revoke-old + partial-failure paths. Closing the loop on AC #3.

PR #1066, opened 2026-05-24
demo(security): DoD acceptance — verify SEMGREP_APP_TOKEN catches AWS keys — DO NOT MERGE

markdbodman:
Final acceptance check for the shift-left security sequence (#1058, #1059, #1060, #1064).

Single deliberate pattern: a canonical AWS access key pair using the official example values (`AKIAIOSFODNN7EXAMPLE`).

**Expected outcome:** Semgrep Scan FAILS with `generic.secrets.security.detected-aws-access-key-id` (or equivalent) on both lines, proving the SEMGREP_APP_TOKEN secret is wired and the full registry is loaded. Will close this PR after capturing the result.

### DIFF
--- a/apps/web/lib/security/__demo-pre-pr-gate.ts
+++ b/apps/web/lib/security/__demo-pre-pr-gate.ts
@@ -1,0 +1,17 @@
+/**
+ * DO NOT MERGE — definition-of-done acceptance demo for the Semgrep gate.
+ *
+ * Verifies that after PR #1064 wired `semgrep ci` to consume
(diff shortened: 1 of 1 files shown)


markdbodman:
Closing — SEMGREP_APP_TOKEN secret not yet set on the repo (workflow log shows empty value). The graceful-degradation path is working (no false failure), but until the secret is added per docs/security/semgrep.md → 'Activating the full ruleset', the gate runs against the limited community ruleset. Re-running this demo after the secret lands will be a 5-min check.

PR #1120, opened 2026-05-24
fix: repair quiescence guard and voice storage CI

markdbodman:
## Summary

- Snapshots the actual `TaskRun.status` in quiescence blockers instead of hardcoding `status: "working"`, so the stall-detection guard stays strict without a new allowlist.
- Preserves Windows absolute voice storage paths when tests or runtime execute on non-Windows hosts, fixing the Linux CI unit failure from PR #1116.

## Verification

- `node scripts/check-no-bare-working-write.mjs`
- `pnpm --filter web exec vitest run lib/voice-synthesis/storage-root.test.ts lib/self-upgrade/quiescence.test.ts`
- `pnpm --filter web typecheck`
- `pnpm --filter web exec next build`

### DIFF
--- a/apps/web/lib/self-upgrade/quiescence.ts
+++ b/apps/web/lib/self-upgrade/quiescence.ts
@@ -564,6 +564,7 @@ export async function captureActiveSessionBlockers(opts?: {
       taskRunId: true,
       title: true,
       buildId: true,
+      status: true,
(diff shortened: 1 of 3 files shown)


markdbodman:
Superseded by #1122, which carries both the voice-storage CI fix and the quiescence guard allowlist fix on top of current main.

PR #1213, opened 2026-05-27
feat: contributor inventory sync — Phases 1-4 (move git/GitHub off request path)

markdbodman:
## Summary

Implements Phases 1-4 of [`BI-063BDF1B`](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1210) (epic `EP-REDUCTION-GEAR-ARCH`) per the spec/plan in [#1210](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1210).

Moves contributor inventory (local git worktrees, remote branches, GitHub PR state) off the per-request shell-out path that PR #1207 shipped. Adds a scheduled Inngest cron (`*/10 * * * *`) that writes a `ContributorInventorySnapshot` Prisma table; the read-model becomes a pure DB query with latest-successful-per-source semantics. The Live portal at `:3000` now renders the dashboard meaningfully with no GitHub credential bound — local-git data populates from the same sync, PR cells show `—` with a "GitHub not connected" header annotation.

## What's in this PR

- **`packages/db/prisma/schema.prisma`** — `ContributorInventorySnapshot` + `ContributorInventorySyncRun` models with FK + ON DELETE CASCADE; `ScheduledJob.metadata` JSON column for durable etag persistence.
- **`packages/db/prisma/migrations/20260526200000_contributor_inventory_snapshot/`** — handwritten reversible SQL.
- **`packages/db/src/seed-contributor-inventory.ts`** — idempotent `ScheduledJob` heartbeat seed for `jobId: "contributor-inventory-sync"`, wired into `seed.ts`.
- **`apps/web/lib/queue/functions/contributor-inventory-sync.ts`** — pure `runContributorInventorySync` runner plus two Inngest function wrappers (`contributorInventorySyncCron` + `contributorInventorySyncOnDemand`) sharing the runner; matches the established pattern from `code-graph-reconcile.ts` and `governed-backlog-tee-up.ts`.
- **`apps/web/lib/contributor-change-lanes/read-model.ts`** — rewritten to read snapshot rows via latest-successful-per-source; `LaneReadModelFreshness` now carries a five-state discriminator (`ok` / `stale` / `error` / `not-configured` / `warming-up`).
[...]

PR #1219, opened 2026-05-27
fix(queue): contributor-inventory-sync heartbeat self-heals on missing row

markdbodman:
## Summary

Phase 7 functional verification of [#1213](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1213) surfaced a fragility in the runner: `prisma.scheduledJob.update()` throws `P2025 RecordNotFound` if the `ScheduledJob` row for `jobId: "contributor-inventory-sync"` doesn't exist. The seed helper `packages/db/src/seed-contributor-inventory.ts` creates the row on install, but upgrade installs that don't re-run `pnpm db seed` would otherwise hit this error on the first cron tick.

Switches to `upsert` with a seed-equivalent `create` payload. The runner is now self-healing.

## What changed

- `apps/web/lib/queue/functions/contributor-inventory-sync.ts` — the heartbeat write at the end of `runContributorInventorySync` becomes an `upsert`. `lastError` and `nextRunAt` are computed once into locals so the `create` and `update` branches can't drift.
- `apps/web/lib/queue/functions/contributor-inventory-sync.test.ts` — 8 existing cases updated to expect the new call shape (single `upsert` call with `where` + `create` + `update`); 2 new cases added:
  - "scheduledJob nextRunAt is set to now + 10 minutes on both create and update branches of the upsert"
  - "heartbeat upsert: self-heals when the ScheduledJob row is missing (P2025 avoided)"

10/10 cases passing locally.

## What did NOT change

Same row, same fields, same write semantics — just an idempotent write instead of a strict update. No data-migration concerns; no schema changes.

## Verification

- **Local vitest** (`pnpm --filter web exec vitest run lib/queue/functions/contributor-inventory-sync.test.ts`): 10/10 pass.
- **Local typecheck:** bypassed at the pre-commit hook because `dev-portal-start` polluted host `node_modules` workspace symlinks (Phase 7 ran the Contributor preview); this is the documented recovery pattern (memory `dev_portal_polluted_node_modules`). CI runs a fresh `pnpm install` and gates merge.

## Related

- Surfaced during Phase 7 verification of [#1213](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1213#issuecomment-4550671238) "finding (B)".
- BI-063BDF1B (already marked `build` triaged, this is a defensive maintenance patch on the same surface).

### DIFF
--- a/apps/web/lib/queue/functions/contributor-inventory-sync.test.ts
+++ b/apps/web/lib/queue/functions/contributor-inventory-sync.test.ts
@@ -21,7 +21,7 @@ const mocks = vi.hoisted(() => ({
   syncRunCreate: vi.fn(),
   syncRunUpdate: vi.fn(),
   snapshotCreateMany: vi.fn(),
-  scheduledJobUpdate: vi.fn(),
+  scheduledJobUpsert: vi.fn(),
 }));
 
 vi.mock("@dpf/db", () => ({
@@ -35,7 +35,7 @@ vi.mock("@dpf/db", () => ({
       createMany: mocks.snapshotCreateMany,
     },
     scheduledJob: {
-      update: mocks.scheduledJobUpdate,
+      upsert: mocks.scheduledJobUpsert,
     },
   },
 }));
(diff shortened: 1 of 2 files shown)


markdbodman:
Closing as duplicate of #1220 which landed 13 seconds later but takes a cleaner approach: the sibling PR re-exports the seed constants (`CONTRIBUTOR_INVENTORY_JOB_ID`, `CONTRIBUTOR_INVENTORY_JOB_NAME`, `CONTRIBUTOR_INVENTORY_SCHEDULE`) from `@dpf/db` so the runner and seed can't drift. My PR inlined the literals to keep the change minimal; deduplication via shared constants is the better long-term choice. Same root finding, same surface; defer to #1220.

PR #1249, opened 2026-05-29
fix(monitoring): stop false Platform Health alerts on macOS/Linux + add platform watch-list

markdbodman:
## Problem

On a first-run macOS install, **Platform Health showed a red `CRITICAL ContainerDown / windows-host`** alert. The base `prometheus.yml` (shared by Windows *and* macOS, since the macOS overlay didn't override it) scrapes the Windows-only `windows_exporter` at `host.docker.internal:9182` — a target that can never come up off-Windows, so the `ContainerDown` rule fired. The same dead job lived in `prometheus.linux.yml`, so a native-Linux install hit the identical false CRITICAL. Separately, the Health Monitoring card showed a permanent false **"Degraded — Host telemetry is not available"** on macOS, where no host exporter exists by design.

## Changes

- **`prometheus.macos.yml` (new) + macOS overlay mounts it** — cross-platform jobs only, no `windows-host`, mirroring how `docker-compose.linux.yml` mounts `prometheus.linux.yml`.
- **`prometheus.linux.yml`** — drop the same dead `windows-host` job (Linux gets host NICs from node-exporter's `node_network_info`).
- **`health-summary.ts`** — `deriveMonitoringSummary` now distinguishes a host-telemetry exporter that's *configured but down* (still Degraded — signal preserved for Windows/Linux) from one that's *not configured on this substrate* (expected → Active). macOS reads Active.
- **`docs/install/platform-support-watchlist.md` (new)** — a living cross-platform gotcha tally, seeded from this fix plus other open items (hardcoded Grafana URL, git-lfs-not-installed, BSD/GNU shell traps, Docker Desktop VM NIC boundary). `AGENTS.md` §2 points to it so it's checked before adding host-coupled config.

## Sweep landmine respected

The macOS/Linux roadmap flags a *"Sweep regression"* risk: the network discovery sweep depends on the `windows-host` scrape **on Windows**. The base `prometheus.yml` keeps the job for Windows; it's removed only where `windows_exporter` can never run (macOS/Linux), where the sweep already degrades to `os.networkInterfaces()` regardless of the Docker Desktop VM boundary. Documented inline in both configs.

## Verification

- Unit tests: 20/20 monitoring tests pass (incl. macOS-absent → Active, Windows-down → Degraded, Linux/macOS configs exclude `windows-host`).
- `tsc --noEmit`: clean.
- Live macOS stack: recreated Prometheus → `windows-host` target gone, alert cleared; real Prometheus `up` data fed through the real summary functions → Platform Status **Operational** and Health Monitoring **Active**.

Covers macOS **and** Linux deliberately, to avoid fixing the same class of bug one platform-PR at a time. Commits are individually revertable.


markdbodman:
Closing in favor of https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1252. #1249 was branched from pre-merge `main` and overlapped the already-merged macOS windows-host fix (#1246/#1248). The replacement contains only the net-new work — the Linux windows-host fix, the macOS degraded-telemetry fix, and the platform support watch-list — cleanly rebased on current `main`. Branch `fix/macos-platform-health-windows-host-false-critical` can be deleted.

PR #1297, opened 2026-05-30
feat(self-upgrade): upgrade window = when the storefront is closed (operating hours)

markdbodman:
## Why
Operators hit `outside-window` at any hour (e.g. 23:00). Root cause: the seed enables self-upgrade but ships **no** `maintenanceWindows`, so `isInMaintenanceWindow` is always false, and there's **no UI to set one** — the scheduled upgrade can never run on its own (BI-300192C2). The "window" was also an allow-list operators mistook for working hours.

## Fix
Derive the upgrade window from the **storefront's operating hours** — upgrades run **whenever the store is closed**. Operators already set operating hours (single source of truth), so there's no separate window to configure and no "no window → never runs" dead-end. Explicit `maintenanceWindows` remain an advanced override.

- **`window.ts`** (pure, tested): `isStoreOpen` (timezone-aware, overnight-safe) + `isUpgradeWindowOpen` (explicit override → else allowed when closed → else allowed, never silently block).
- **`operating-hours-read.ts`**: no-auth system reader so the cron path reads the schedule without a user session.
- **`runSelfUpgrade` + `triggerSelfUpgrade`** gates now use the operating-hours model (force/dryRun still bypass).
- **`getSelfUpgradeStatus`** reports `inMaintenanceWindow` (upgrades may run now), `windowConfigured=true`, `windowSource`, `storeOpen` — removing the false "no maintenance window configured" state.

## Verification
- `window.test.ts`: **12** (timezone shifts + the exact 23:00 case + overnight + explicit override).
- Orchestrator + promotions gate tests rewired to the operating-hours model. Full affected suite **279 passing**; `tsc` clean; `next build` exit 0. No migration.

## Follow-ups (filed)
- BI-EF743966 — make `checkIntervalHours` real (still inert; separate concern).
- BI-F0E4272B — manual "Trigger Now" shouldn't require a separate Emergency override.
- An explicit-window editor (BI-300192C2) is now optional since the window derives from operating hours.


markdbodman:
Now also makes `checkIntervalHours` real (BI-EF743966): the scheduled cron records `lastCheckedAt` and skips ticks until the configured interval has elapsed, so the number actually governs poll cadence (manual/forced runs are never throttled but reset the clock). Pure `isCheckIntervalElapsed` + 6 unit tests + orchestrator throttle cases. Branch synced with main; typecheck + self-upgrade suite green.

PR #1520, opened 2026-06-06
fix(self-upgrade): skip before drain + cooldown to stop the quiescence loop

markdbodman:
## Problem

On the live install, self-upgrade was forcing the portal into `level:"draining"` on a tight ~6-minute cycle — every new server action / MCP call refused with `QuiescingError` (`PORTAL_QUIESCING`, retryAfter 30s). The portal was unusable ~5 of every 6 minutes during concurrent dev sessions.

Evidence from the running install:
- Every recent `QuiescenceRun` was `trigger='self-upgrade'`, status `deferred`/`failed`, with **empty `targetBundleHash`/`targetVersion`**.
- **No promoter container/image** present (`promoterImage: dpf-promoter` was never built), yet the portal kept draining for the full quiescence budget, then failing at `docker run` and re-triggering seconds later.
- The portal container itself never recycled and Inngest/coordinator were healthy — so this was the orchestrator looping, not an infra outage.

## Root causes

1. `runSelfUpgrade` created a run and entered the quiescence **drain before checking whether a swap was even possible** — no promoter-image check, no nothing-newer check. `startQuiescence` was never passed a target, so every drain was recorded against an empty bundle.
2. After a deferred/failed run there was **no backoff**, so the next trigger (scheduled cron, manual `ops/self-upgrade.run` event, or autonomous deploy) re-drained the portal within seconds.

## Fix

**Skip before drain**
- `promoter.ts` `isPromoterAvailable()` — `docker image inspect` precheck. Absent image / unreachable docker → skip **before** any drain (`reason: "promoter-unavailable"`), instead of draining then failing at `docker run`.
- `runSelfUpgrade` nothing-newer guard — if the honest built stamp already equals the deployed SHA, skip before drain (`reason: "up-to-date"`).
- `startQuiescence` is now stamped with the real `targetVersion`/`targetBundleHash`, so a drain is never recorded against an empty bundle.

**Cooldown backoff** (`self-upgrade/cooldown.ts`, new)
- PlatformConfig-backed `getCooldownUntil` / `isInCooldown` / `recordCooldown` / `clearCooldown`. A gate at the top of `runSelfUpgrade` skips any non-`force`, non-`dryRun` trigger while a cooldown is active — breaking the re-drain loop **regardless of trigger source**.
- Cooldown is set on every defer/fail (prep conflict, quiescence defer/abort/fail, recovery-point failure, promoter spawn error, promoter non-zero) and **cleared on a successful swap**. Configurable via `self_upgrade.cooldownMinutes` (default 30).

`force`/`dryRun` bypass both guards and the cooldown (operator asked now / dry-run never swaps). A skip never drains and sets no cooldown, so the upgrade resumes the moment a real bundle + promoter are available and the window is open.

## Acceptance criteria

- [x] No newer bundle / no promoter available → **no drain**, no `QuiescenceRun` with empty `targetBundleHash`.
- [x] After a deferred/failed run → cooldown before the next attempt (default 30 min); portal not re-drained within minutes.
- [x] During an active session the portal stays `normal` unless a real bundle exists (nothing-newer guard) and the maintenance window is open (existing scheduled-window gate).
- [x] Regression tests for skip-before-drain (no target / no promoter) and cooldown-after-defer.

## Tests

- `cooldown.test.ts` — pure `isInCooldown` gate.
- `self-upgrade.test.ts` — skip-before-drain (promoter-unavailable, nothing-newer, target stamping) + cooldown (skip while cooling down incl. event-triggered, set on defer/fail, cleared on success, force/dryRun bypass).
- `117` self-upgrade unit tests pass; `301` pass across the self-upgrade surface. Local `next typegen && tsc` is blocked by this worktree's incomplete `node_modules` (`next` not installed / `@dpf/db` unresolved repo-wide) — typecheck is left to CI.

Live manifestation of the **BI-A3930CD7** ("defer during active sessions") / **BI-CAPSULE-SELFUPGRADE-001** cluster. No overlap with open PRs.


markdbodman:
### Added: "Upgrade activity" panel on /ops/self-upgrade

Per follow-up request — surface the blocking activities so operators can see what's happening during a self-upgrade instead of silent `PORTAL_QUIESCING` refusals.

- `quiescence.ts` `getQuiescenceActivity()` reads the live quiescence level + the active/most-recent `QuiescenceRun` and collapses its `ActiveSessionBlockers` snapshot into one friendly line per surface (label + count + worst-case wait).
- `getSelfUpgradeStatus()` now returns `quiescence` (level, target, deferSurface, blockers) + `cooldownUntil`.
- `SelfUpgradeClient` renders the panel when draining/swapping, when a drain just deferred, or while a post-defer/fail cooldown is active — e.g. **"Waiting on: Build Studio — build phase / AI coworker working ×6 · ~30s"**, the target build, and **"Automatic upgrades paused until &lt;time&gt;"**. It states plainly that deferred work was never interrupted, and stays hidden when the level is normal with no cooldown.

All 19 CI checks green (Typecheck, Unit Tests, Production Build included).

PR #1581, opened 2026-06-06
fix(build): restore main Unit Tests gate — downgrade-gate fabrication retries + update #1572/#1573 tests

markdbodman:
## Problem
`main`'s **Unit Tests gate is red** — 6 tests failed because #1572 (fabrication-retry) and #1573 (bounded review) merged **without updating their tests** (the local suite was skipped). A red required gate blocks clean merge of *every* PR.

## Fixes
**1. Correctness improvement to #1572.** It made *all* build-route fabrication retry 3×, including a **healthy** false claim — but a healthy false claim is genuine fabrication, not a transient blip, so retrying it just wastes calls and delays the guard. Gate the extra retries on `result.downgraded`: a **downgraded** build-route turn gets 3 attempts (ride out a 529 blip); **healthy** + conversational keep one retry so the guard fires promptly. This matches what the agentic-loop tests encode.

**2. Tests updated to shipped behaviour:**
- `agentic-loop.test.ts` — the downgraded build-route test now exhausts 3 retries (4 fabricated claims); the 3 healthy tests are restored unchanged by the gate.
- `build-reviewers.test.ts` — assert the new bounded **DECISION DISCIPLINE / SIZE-AWARENESS** contract instead of the removed "report ALL issues / ZERO surprise".
- `build-disciplines-integration.test.ts` — match the new uppercase `TEST-FIRST`/`BITE-SIZED`.

Restores the Unit Tests gate so #1576/#1577 (and all PRs) can merge cleanly. (vitest isn't installable in the worktree; relying on CI as the binding evidence, with each retry-count reasoned.)

### DIFF
--- a/apps/web/lib/integrate/build-disciplines-integration.test.ts
+++ b/apps/web/lib/integrate/build-disciplines-integration.test.ts
@@ -241,8 +241,8 @@ describe("Build Disciplines — Full Flow Integration", () => {
         fileStructure: [{ path: "test.ts", action: "create", purpose: "test" }],
         tasks: [{ title: "task 1", testFirst: "write test", implement: "code", verify: "run" }],
       });
-      expect(prompt).toContain("test-first");
-      expect(prompt).toContain("bite-sized");
+      expect(prompt).toContain("TEST-FIRST");
+      expect(prompt).toContain("BITE-SIZED");
     });
   });
 });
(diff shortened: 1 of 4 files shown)


markdbodman:
Superseded by #1579 (which greened the same 6 tests after #1572/#1573). main is green and has all fixes. The only unique change here was a minor correctness nicety — gating the build-route fabrication retries to downgraded turns so a *healthy* false claim fires the guard promptly instead of retrying 3x. Filing that as an optional follow-up rather than re-conflicting with #1579's test updates.

PR #1638, opened 2026-06-07
ci: cache Next.js build to speed the Production Build gate

markdbodman:
## Why

Follow-up to #1632. With unit tests now sharded to ~2 min, the new CI critical path is **Production Build (~3m26s)** — the single slowest job and a *required* merge gate (`Typecheck` / `Production Build` / `DCO`). It ran a cold `next build` on every PR, recompiling every module from scratch.

## What

Restore Next.js's `.next/cache` (SWC/webpack compilation cache) via `actions/cache@v4`, keyed on `pnpm-lock.yaml` + source hash, with a deps-only `restore-keys` fallback so a typical PR warm-starts from the most recent build.

- Coverage-neutral and output-neutral — only recompilation of *unchanged* modules is skipped; the produced build is identical.
- Per [Next.js CI caching guidance](https://nextjs.org/docs/app/building-your-application/deploying/ci-build-caching).
- The win lands on the **second run onward** — the first run seeds the cache (expect no change, possibly +a few seconds for the cache save).

## Verification

CI on this PR seeds the cache; a re-run (or the next PR) should show Production Build drop. I'll report the before/after once a warm run lands.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -259,6 +259,20 @@ jobs:
       - name: Generate Prisma client
         run: pnpm --filter @dpf/db exec prisma generate
 
+      # Restore Next.js's build cache (SWC/webpack compilation artifacts). On a
+      # cache hit, unchanged modules are not recompiled, which is the bulk of a
+      # cold `next build`. Key on the lockfile + all source so any source change
+      # rolls the key; restore-keys fall back to the most recent build for the
+      # same deps so a typical PR still warm-starts. See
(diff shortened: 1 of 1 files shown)


markdbodman:
Closing — verified ineffective. Warm re-run got a cache hit but restored only 358 KB and the build stayed at ~181s. Root cause: this app builds with **Turbopack** (Next.js 16 default), which does not use `.next/cache` for its compilation cache, so directory-caching it saves nothing. The webpack-era `.next/cache` pattern doesn't apply here. No code merged; CI stays as-is from #1632.

PR #1804, opened 2026-06-13
feat(workbooks): persistent grid views (EP-GRID-WORKBOOKS Phase 3)

markdbodman:
A grid's view now **sticks**: the quick filter, per-column filters, sort, conditional-format rules, and the provenance toggle are saved per tableId and restored on reload. Client-side (localStorage) so it works for every grid — custom and platform — with **no migration**.

> Stacked on #1803 (chart) → #1783 (umbrella). Base auto-retargets as the stack merges.

Serialization is a pure, unit-tested `grid-view-state.ts` with **defensive parsing**: bad JSON → null; wrong-typed fields dropped; invalid conditional-format rules and sort entries filtered — a stale/corrupt entry never breaks a grid load. The Grid hydrates once per tableId then saves on change.

Named, shareable server-side views (persisting to the existing `WorkbookView` model) are a tracked follow-up.

Gate: grid-view-state vitest 6 passing; web typecheck + production build green.

### DIFF
--- a/apps/web/components/workbooks/Grid.tsx
+++ b/apps/web/components/workbooks/Grid.tsx
@@ -6,7 +6,14 @@
 // implementation detail contained here, so it can be swapped without touching
 // the data layer, server, or pages.
 
-import { useCallback, useMemo, useRef, useState, type KeyboardEvent as ReactKeyboardEvent } from "react";
+import {
+  useCallback,
+  useEffect,
+  useMemo,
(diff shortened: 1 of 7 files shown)


markdbodman:
Folding this into the umbrella #1783 (consolidated to a clean signed linear history to fix the DCO/merge-commit issue). Content is preserved in #1783.

PR #1865, opened 2026-06-14
feat(finance): invoice tax default, SMTP pre-flight, e-signature (Phase 1)

markdbodman:
Closes the three invoice gaps from the Runs 6 & 7 fresh-install archetype audit (`docs/testing/handoff-invoice-gaps.md`). Worked directly in a worktree (Build Studio is down); functional verification is deferred to the mass end-to-end test plan, so this PR focuses on correct implementation + unit tests + green CI.

## Gap 1 — Tax default not wired to the org's VAT selection (bug)
The new-invoice form hardcoded `taxRate: 20` on every line regardless of the wizard VAT choice. Now resolved from `OrganizationTaxProfile.taxModel` + the applied finance profile's `defaultTaxRate` (**0 for No-VAT orgs**, the profile's per-industry rate for VAT-registered orgs — correctly 0 for VAT-exempt industries like healthcare). Pure, unit-tested resolver in `lib/finance/invoice-default-tax.ts`, passed to `CreateInvoiceForm` as a prop — mirroring how bills already default.

## Gap 2 — Send Invoice silently fails with no SMTP (bug)
On a cold-start install with no SMTP, `sendEmail` fake-succeeded in dev (silent no-op) or threw a generic 500 in prod. The send route now pre-flights `isEmailConfigured()` and returns a **422 with an actionable message before any state mutation**, so a failed send never leaves the invoice falsely marked `sent`. The existing `InvoiceSendButton` already surfaces `body.message`.

> Note: there is **no SMTP settings UI** today (SMTP is env-var configured), so the message is truthful about that rather than linking a non-existent page. Surfacing an SMTP config UI/onboarding prompt is a separate follow-up.

## Gap 3 — E-signature (Phase 1, feature)
In-platform signature capture for invoices:
- **Schema + migration**: `Invoice.signatureRequired / signedAt / signedByName / signedByEmail / signatureDataUrl` (additive).
- **Create form**: "Require signature before payment" toggle, defaulted **on for `legal-services` / `accounting`** archetypes, off elsewhere (operator can still enable per invoice).
- **Public payment portal**: a canvas signature pad gates the Pay Now block until the customer signs; shows a "Signed by … on …" confirmation once signed.
- **On sign**: records the signature and sends a best-effort signed-confirmation email with the **countersigned PDF** (skipped cleanly when SMTP is unconfigured).
- **Admin invoice detail**: signature status + an operator toggle to require/remove signature before signing.

Phase 2 (third-party DocuSign/HelloSign for standalone documents) is **out of scope** here — to be filed as its own BI.

## Verification
- ✅ **67 unit tests pass** (`vitest`) across the 4 affected files: tax resolver, archetype signature default, `isEmailConfigured`, and `signInvoice` / `setInvoiceSignatureRequired` server actions (auth, idempotency, validation, rejections).
- ✅ **schema-regression-guard** and **bundle-boundary-guard** pass locally.
- ✅ `prisma validate` + `prisma generate` succeed with the new fields.
- ⚠️ The local pre-commit `tsc` was skipped (`DPF_SKIP_TYPECHECK=1`): the deps-less worktree resolves `@dpf/db` to the root clone's pre-change Prisma client, so the only local type errors are the five new fields (proven by triage) plus pre-existing `@dpf/storefront-templates` resolution noise in untouched files. **CI's typecheck job regenerates the client from this schema first and is the binding check.**

Functional happy-path verification (create → send → sign → pay) will be exercised in the mass E2E test plan.


markdbodman:
**CI status: all checks for this PR's changes are green.** Typecheck ✅ · Production Build ✅ · Unit Tests (packages) ✅ · web shards 1/2/4 ✅.

The only red is **Unit Tests (web shard 3/4)**, whose sole failing file is `lib/inference/ollama-health.test.ts` (`checkBundledProviders`). This is **inherited main breakage, unrelated to this PR** — it was broken on `main` by #1860 (which merged after this branch was cut) and is fixed by the still-open #1863. This PR touches **zero** inference files (only finance/email/invoice/PDF/schema). Because PR CI runs against the branch *merged with current main*, shard 3 will go green automatically once #1863 lands; re-running CI after that merge should produce an all-green run with no change here.

PR #1877, opened 2026-06-14
fix(finance): default recurring-invoice & PO line tax from org VAT (not 20%)

markdbodman:
Completes the Runs 6 & 7 tax-default gap. [PR #1865](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/1865) fixed the hardcoded 20% TAX default on the one-off invoice form, but a thoroughness sweep found **two more finance line-item forms with the same `taxRate: 20` hardcode**:

| Form | Type | Before | After |
|---|---|---|---|
| `CreateInvoiceForm` | AR (invoice) | ✅ fixed in #1865 | from org VAT |
| `CreateBillForm` | AP (bill) | ✅ already 0% | 0% |
| **`CreateRecurringForm`** | AR (recurring invoices) | ❌ hardcoded 20 | **from org VAT** |
| **`CreatePOForm`** | AP (purchase orders) | ❌ hardcoded 20 | **0% (mirrors bills)** |

## Changes
- Extracts **`getInvoiceDefaultTaxRate()`** into `lib/actions/financial-setup.ts` — one DB→resolver path shared by the invoice and recurring-schedule forms (the invoice page is refactored onto it, so the logic can't drift). Recurring invoices bill customers, so they default tax the same way one-off invoices do.
- `CreatePOForm` gains a `defaultTaxRate` prop defaulted to **0** (`DEFAULT_PO_TAX_RATE`), matching the accounts-payable decision already in `bills/new` (#1759).
- Unit tests for `getInvoiceDefaultTaxRate` (vat → rate, none → 0, no-profile bootstrap fallbacks, empty → 0).

## Verification
- ✅ 28 unit tests pass (`financial-setup.test.ts` + `invoice-default-tax.test.ts`).
- ✅ No type errors in the changed files (CI Typecheck is binding — deps-less worktree resolves `@dpf/db` to the root clone's client, so the only local errors are unrelated pre-existing env noise in untouched files).
- Functional verification of the TAX % pre-fill on these forms is covered by the audit plan's **Phase G-REG-1** (I'll extend it to name the recurring and PO forms too).

### DIFF
--- a/apps/web/app/(shell)/finance/invoices/new/page.tsx
+++ b/apps/web/app/(shell)/finance/invoices/new/page.tsx
@@ -1,13 +1,14 @@
 // apps/web/app/(shell)/finance/invoices/new/page.tsx
 import { prisma } from "@dpf/db";
-import { getFinancialProfile } from "@dpf/finance-templates";
 import Link from "next/link";
 import { CreateInvoiceForm } from "@/components/finance/CreateInvoiceForm";
-import { resolveInvoiceDefaultTaxRate } from "@/lib/finance/invoice-default-tax";
+import { getInvoiceDefaultTaxRate } from "@/lib/actions/financial-setup";
 import { defaultSignatureRequiredForArchetype } from "@/lib/finance/invoice-signature-default";
 
 export default async function NewInvoicePage() {
-  const [customers, taxProfile, orgSettings, storefront] = await Promise.all([
+  // Default the TAX % field from the org's wizard VAT selection, not a 20% hardcode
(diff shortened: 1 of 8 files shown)


markdbodman:
Update: the G-REG-1 audit-plan extension (covering `/finance/recurring/new` and `/finance/purchase-orders/new` alongside the invoice + bill forms) is **included in this PR** — so the code fix and its regression check land together. Phase G-REG itself is already on `main` via #1874.

PR #1895, opened 2026-06-14
feat(workbooks): surfacing block + grid on 8 surfaces (finance, customer, portfolio, people, compliance×3)

markdbodman:
## What

The **registry-driven building block** for making the grid an alternative form on *every* multi-item surface — the operator-chosen "systematic wrapper first" path over per-surface hand-wiring.

- **`PlatformGridSection({ entityType, view })`** = `SurfaceViewSwitcher` + (`view ? SurfacePlatformGrid : null`) in one component. Composes the existing pieces; no new grid code.
- **`parseSurfaceView(raw)`** (`lib/workbooks/surface-view.ts`) — pure, unit-tested helper narrowing `?view=` to grid/board/null, replacing the copy-pasted inline check on every page.
- **Adopted on `/finance/invoices`** as the proof — behavior unchanged: switcher + grid via the block, list rendered under `{!view && …}`.

## Why

The grid engine is complete but embedded on only ~3 surfaces via ~8 lines of hand-wiring each. This block turns adding the grid to a new surface into **~3 lines** (parse view + one component + wrap the list), and keeps the switcher/grid wiring in one place so it stays consistent and can't drift per page.

## Rollout after this

Remaining registered entities (`digital_product`→portfolio, `customer_account`→customer, `employee_profile`→people) and then the unregistered list surfaces (compliance, CRM, finance sub-lists, inventory, rental, storefront) adopt this block + a `PLATFORM_TABLES` config entry — entity by entity, each its own small PR.

Gate: `vitest` (parseSurfaceView grid/board/null/unknown); `pnpm --filter web typecheck` green.

### DIFF
--- a/apps/web/app/(shell)/compliance/controls/page.tsx
+++ b/apps/web/app/(shell)/compliance/controls/page.tsx
@@ -1,12 +1,14 @@
 import { listControls } from "@/lib/actions/compliance";
 import { CreateControlForm } from "@/components/compliance/CreateControlForm";
 import { FilterBar, StatusBadge } from "@/components/ui/report-kit";
+import { PlatformGridSection, parseSurfaceView } from "@/components/workbooks/PlatformGridSection";
 import Link from "next/link";
 
-type Props = { searchParams: Promise<{ controlType?: string; implementationStatus?: string; effectiveness?: string }> };
+type Props = { searchParams: Promise<{ controlType?: string; implementationStatus?: string; effectiveness?: string; view?: string }> };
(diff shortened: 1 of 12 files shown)


markdbodman:
Added: Customers surface (/customer) now uses PlatformGridSection — first new-coverage adoption (account list → List/Grid/Board). Note recorded in-PR: /employee needs a small tweak before wiring (its existing ?view= tab param collides with the grid's ?view=). typecheck green.

PR #2383, opened 2026-06-25
fix(build-studio): preserve decomposition override across review re-runs

markdbodman:
## What

Fixes the broken decompose-override → advance path in Build Studio. When a design trips the Phase-4b `decompose-required` gate, an operator override (`record_decomposition_override`) is meant to let the build proceed — but `reviewDesignDoc` rewrote `designReview = {...review, sizeAssessment, reviewers}` without preserving `decompositionOverride`, so the gate (and `resume-pre-build-phase`, which re-runs reviewDesignDoc to advance overridden builds) saw `!hasOverride` and re-parked them forever. This re-attaches the prior override into the write.

## Why — overnight drain diagnosis (2026-06-25)

The backlog drain was completing **zero** builds. Root-caused to three compounding issues:

1. **Ideate heartbeat = 90s** (the #1 wall) — robust Codex design calls take ~3 min and were reaped every time. **FIXED LIVE at runtime** (`BuildStudioStallThreshold` phase.ideate 90→600s; reversible, no deploy). Proven: 4 builds then cleared ideate and produced designDocs, vs zero before.
2. **The decompose-override → advance path was broken** — **this PR.**
3. **The decompose size-assessment false-positives** — `size-design-doc.ts` counts PascalCase prose mentions as "models" (a docs task scored "8 models → xlarge"). Follow-up calibration fix (count only `dataModel`-field models, or exempt doc/chore work); left for review since it touches the decompose policy (WWMD recommends `auto-decompose` for genuinely-big designs).

## Validation

- Pre-commit scoped typecheck + full `tsc --noEmit`: clean.
- Not yet validated live — needs a portal deploy. 4 builds are parked at the decompose gate and will advance once this ships and an override is recorded.

### DIFF
--- a/apps/web/lib/build/size-design-doc.test.ts
+++ b/apps/web/lib/build/size-design-doc.test.ts
@@ -262,6 +262,30 @@ describe("sizeDesignDoc — heuristic edge cases", () => {
     expect(result.breakdown.models.count).toBe(1);
   });
 
+  it("does NOT count PascalCase prose mentions as models (docs false-positive regression)", () => {
+    // Regression for FB-4227B23B: a docs-only design (author docs/install/windows.md)
+    // that merely *mentions* component/product names in prose was scored as
+    // "8 models -> decompose-required" and parked at the Phase-4b gate. Models are
+    // counted only from the structured dataModel field; prose PascalCase is a
+    // reference, not a new data model.
+    const doc: BuildDesignDoc = {
+      problemStatement:
+        "The WindowsInstaller and DockerDesktop setup for DpfPlatform is undocumented; the InstallGuide must cover ModelRunner and QdrantStore.",
+      dataModel: "",
+      reusePlan: "Reuse the existing MacOsGuide and LinuxGuide structure.",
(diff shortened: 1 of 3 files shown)


markdbodman:
**Update — #3 (the decompose size false-positive) is now fixed in this PR too.**

- Count decompose-size models only from the structured `dataModel` field; prose mentions in `proposedApproach`/`problemStatement` no longer false-count as data models (commit `4cd9d40`).
- **21/21 `size-design-doc` tests green**, including a new docs-false-positive regression: a prose-heavy docs design now scores 0 models → `ok` (was 8 → `decompose-required`). Every existing fixture already declared its models in `dataModel`, so no true-positive is lost.

So this PR now addresses **both** the override→advance path (#2) and the size false-positive (#3). With the 90s→600s ideate heartbeat (#1) already live at runtime, **merging + deploying this PR unblocks the full build pipeline** end-to-end. The size-counting narrowing is still worth a calibration glance, but it's tested and conservative.

PR #2393, opened 2026-06-25
refactor(mcp): third tool pack — work-capsules (lazy) (BI-ARCH-TOOLPACKS)

markdbodman:
Third domain pack (second lazy one), continuing the toolpacks split here.

> **Stacks on #2392** (runtime pack). Merge #2392 first; this then rebases to just the work-capsules diff.

## What
- **`lib/mcp/packs/work-capsules-pack.ts`** — the 10 Work Capsule tools. Definitions moved **verbatim** out of the inline `PLATFORM_TOOLS` array; each handler is a thin wrapper that lazy-imports `@/lib/work-capsules/mcp-handlers` only when called, passing the same arguments the switch case did (2 read tools take params only; 8 write tools take params + userId + context). Grants mirror `TOOL_TO_GRANTS`.
- **`mcp-tools.ts`** — register the pack; remove the inline definitions, the 10 switch cases, and the now-unused `WORK_CAPSULE_TOOL_ENUMS` + its import. Compose back through the registry; dispatch via `registry.getHandler` (after the kernel gate). **Zero behavioral change.**
- **`tool-registry.test.ts`** — work-capsules parity (handler-per-def, grants no-drift, all present in `PLATFORM_TOOLS`) + three-pack composition.
- **`docs/architecture/mcp-tool-packs.md`** updated.

## Verification (web worktree, `prisma generate`)
- `pnpm --filter web typecheck` — **clean**
- `tool-registry` (13) + `mcp-tools-work-capsules` + `mcp-tools` (34) + `mcp-governed-execute` — **74 tests pass**

Three packs down (deliberation+SIEM static; runtime + work-capsules lazy). Remaining: workbooks, then the inline-handler clusters (backlog/build-evidence, sandbox, wiki).

EP-PLATFORM-CONSOLIDATION · spec §6.2

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2396, opened 2026-06-25
fix(ea): clear Style Drift Guard false-positive on a PR-ref comment (unblocks merge queue)

markdbodman:
**main is currently red on the Style Drift Guard**, which blocks every PR branched off it (the merge queue treats the guard as required).

## Root cause
`apps/web/lib/ea/reconcile-sysml-projections.ts` has a code comment referencing `#2385`. The Style Drift Guard's hex-color regex (`#[0-9a-fA-F]{3,8}`) matches it — **every `#NNNN` PR reference is valid hex** — so the guard counted it as a *new hardcoded hex* and failed. The comment landed on main via #2372, so the guard has been red on `origin/main` itself.

## Fix
Reword the reference `#2385` → `PR 2385` in the comment. **Comment-only, zero behavior change.** Guard passes locally (`Style drift OK — 257 files / 1353 known hex`).

## Follow-up (not in this PR)
`scripts/check-style-drift.mjs` should ignore hex inside comments / `#NNNN` PR refs so this can't recur on the next PR-number reference.


markdbodman:
Superseded — main already carries an equivalent fix. #2390 (`154beb910`) reworded the same comment `#2385` → `PR-2385` on `reconcile-sysml-projections.ts`, so the Style Drift Guard already passes on `origin/main`. The tool-pack stack (#2393–#2395), now rebased onto current main, inherits that fix and is green. Closing as redundant.

PR #2403, opened 2026-06-26
fix(golden-triangle): triangle corners snap to their preset (was reading as Custom)

markdbodman:
**Founder report:** pushing the dot to a corner reads as **"Custom"**, but clicking the preset that sits in that same corner shows the preset's name + description. *"The extreme movement control should be the same as the default for that same position."*

**Cause:** presets sit at `0.8`-dominant positions, but the triangle's corners are the `1.0` extremes, and every drag/keyboard/input emitted `preset: "custom"` — so a corner never resolved to the preset that owns it.

**Fix:** a pure `snapToPreset(weights)` — a clearly dominant axis (≥ 0.6) snaps to that axis's preset, a near-centroid posture snaps to **Balanced**, the in-between stays a custom fine-tune. Drag, keyboard, and the number inputs all route through one snap-aware `emit()`, so a corner now reads exactly like clicking its preset button (same label + description).

**Heads-up (separate decision):** the triangle's heaviest reachable posture is now **Assured** (deep review). The extra **"debate"** rung only ever triggered by dragging to an extreme custom quality (≥ 0.85), so it's no longer drag-reachable. If debate should be first-class, it wants its own preset — flagged, not silently dropped.

**24 posture-display tests (5 new) + 5 control tests green; web `tsc` clean.**

### DIFF
--- a/apps/web/components/golden-triangle/GoldenTriangleControl.tsx
+++ b/apps/web/components/golden-triangle/GoldenTriangleControl.tsx
@@ -18,6 +18,7 @@ import {
   decodePostureForDisplay,
   pointToWeights,
   preferenceFromPreset,
+  snapToPreset,
   weightsToPoint,
 } from "./posture-display";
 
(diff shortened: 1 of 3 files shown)


markdbodman:
Superseding the snap approach. Founder clarified: the MAX of a dimension should mean its full setting (Quality-max = debate), with a plain explanation of what min/maxing includes — the process scales continuously with the triangle. Snapping the corner to the Assured preset would cap Quality at review and block the debate max, so it's the wrong move. Reworking as: keep the triangle continuous (the compiler already reaches debate at extreme Quality), give every position a meaningful label (not 'Custom'), and add an axis explainer. New PR incoming.

PR #2463, opened 2026-06-26
fix(build-studio): catch QuiescingError in fire-and-forget startBuildPhaseRun during quiescence drain (BI-QUIESCE-005)

markdbodman:
## Problem

`startBuildPhaseRun` throws `QuiescingError` **outside** its `try/catch` — the BI-QUIESCE-005 entry-point gate that refuses new phase transitions while the portal is draining for a self-upgrade. The four fire-and-forget `void startBuildPhaseRun(...)` callers had **no `.catch`**, so during a drain the rejection escaped as a **real unhandled promise rejection in the production Node/Next.js server process** — not just a test-teardown artifact.

This is the production-side companion to the earlier test-determinism fix that mocked these background modules in `build-phase-run.test.ts` / `build-governed.test.ts`.

## Fix

Append `.catch(() => {})` to all four fire-and-forget sites:

- `apps/web/lib/actions/build.ts` — `createFeatureBuild` (ideate-phase start)
- `apps/web/lib/mcp-tools.ts` ×3 — the `ideate→plan` (fix + standard) and `plan→build` phase-advance paths

Cost tracking is deliberately non-blocking, so a refused start during a drain is **expected and silent** — swallowing the `QuiescingError` is the correct behavior. The companion `void completeBuildPhaseRun(...)` calls are intentionally left untouched: that function wraps everything in `try/catch` and never throws.

**Reusable gotcha:** every quiescence entry-point gate (`spawnWorkThread`, `sandboxPool.acquire`, `callBrowserUse`, `startBuildPhaseRun`) throws `QuiescingError` *outside* its `try/catch`, so any fire-and-forget `void` caller of one needs a `.catch`.

## Test

New regression test in `build-governed.test.ts`: mocks `getQuiescenceLevel → "draining"`, captures `process.on('unhandledRejection')`, and asserts `createFeatureBuild` resolves with **no** unhandled rejection (and that the drain gate refuses before any DB write).

## Verification

- **With fix:** 27/27 pass (`build-governed` 20 + `build-phase-run` 7), re-confirmed against the rebased `origin/main` tree.
- **Without fix (temp-reverted):** the new test **fails**, capturing the exact `QuiescingError { level: "draining" }` — proving it's a true regression lock, not a tautology.
- **Typecheck:** the three changed files are clean (verified via project `tsc`, filtering unrelated stale-client drift noise). CI runs the authoritative typecheck.

### DIFF
--- a/apps/web/lib/actions/build-governed.test.ts
+++ b/apps/web/lib/actions/build-governed.test.ts
@@ -9,6 +9,9 @@ const { mockAuth, mockPrisma } = vi.hoisted(() => ({
       update: vi.fn(),
       create: vi.fn(),
     },
+    buildPhaseRun: {
+      upsert: vi.fn(),
+    },
     businessBuildBrief: {
       findUnique: vi.fn(),
       upsert: vi.fn(),
@@ -79,6 +82,10 @@ const { mockRunBuildPipeline } = vi.hoisted(() => ({
   mockRunBuildPipeline: vi.fn(),
 }));
 
+const { mockGetQuiescenceLevel } = vi.hoisted(() => ({
+  mockGetQuiescenceLevel: vi.fn(),
+}));
+
 vi.mock("@/lib/auth", () => ({
   auth: mockAuth,
 }));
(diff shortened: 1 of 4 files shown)


markdbodman:
Note for reviewers: the **Module Size Guard** check is red due to `apps/web/components/ops/SelfUpgradeClient.tsx` (1217 → 1221), which **this PR does not touch** — it is pre-existing drift on `main` (origin/main fails the same guard independently). This PR is module-size clean: it grows no file past baseline, and surgically re-baselines only its own regression-test file (`build-governed.test.ts` 1051 → 1114) via the guard's documented intentional-growth path. The guard is advisory (non-blocking).

PR #2578, opened 2026-07-04
fix(inference): back off CLI pool when rate-limit has no parseable Retry-After

markdbodman:
## Symptom
The portal on :3000 intermittently stops responding — "happening a lot." No crash, no OOM (`oom_kill 0`), `RestartCount=0`, healthy between incidents. The unresponsiveness is **transient event-loop / Docker-daemon starvation** during a **Codex CLI retry-storm** driven by Build Studio.

Observed on the live install (2026-07-04): codex rate-limit + child-timeout events peaked at **87/hr** (08:00 UTC), portal sluggish during the window, self-recovered when the storm cleared.

## Root cause (confirmed against live DB + logs)
Build Studio fans out agents (e.g. `reviewDesignDoc` = 3 parallel reviewers). Each shells out a `docker exec dpf-sandbox-1 codex` child held open up to 10 min (`CLI_TIMEOUT_MS`) through the single shared sandbox. OpenAI rate-limits the codex subscription.

There **is** a backoff gate: `ai-inference.ts` consults `getCliPoolStatus` before dispatching a CLI-backed call and fast-fails to fallback when the pool `isExhausted`. **But the gate was a permanent no-op for codex:**
- `isExhausted` requires `resetAt` in the future.
- `recordCliRateLimit` only set `resetAt` when `parseRetryAfterSeconds` returned a value.
- Codex's rate-limit output (`"Reading prompt from stdin... OpenAI Codex v0.137.0 ..."`) carries **no parseable Retry-After** → `resetAt` stayed `null` → `isExhausted` always `false` → gate never tripped.

Live `CliPoolStatus` row confirmed: `resetAt = null`, `retryAfterSeconds = null`. So every agent kept spawning doomed 30s-retry children into an already-rate-limited pool — the thundering herd that hung the portal.

## Fix
- **Always back off on a recorded CLI rate limit**: use the provider's Retry-After when present, otherwise a bounded default window (`DEFAULT_CLI_COOLDOWN_SECONDS`, 60s, env-overridable via `DPF_CLI_RATELIMIT_DEFAULT_COOLDOWN_SECONDS`), capped at `MAX_CLI_COOLDOWN_SECONDS` (1h) so a misparse or long subscription reset can't wedge a provider off the fallback chain. A successful call still calls `clearCliRateLimit`, so the window is only ever an upper bound.
- **Teach `parseRetryAfterSeconds` the hour/minute reset formats** ChatGPT/Codex subscription limits actually emit ("try again in 2h 30m").
- `retryAfterSeconds` still records only what was parsed (may be null) for observability; `resetAt` is what the gate reads.

Net effect: a rate-limited codex now backs off ~60s and dispatches fall back to the next provider, instead of storming the shared sandbox and the portal.

## Tests
`apps/web/lib/routing/cli-pool-status.test.ts` — 19/19 pass. Updated the two tests that encoded the old "null resetAt = not blocking" behavior; added coverage for the default-cooldown backoff, the cap, and hour/minute parsing.

## Follow-up (not in this PR)
The codex-cli adapter still has **no concurrency cap** (the local adapter has `withLocalInferenceLock`; codex-cli has nothing), so the shared sandbox can still be hit by unbounded parallel children. Filing a separate BI to serialize/limit codex-cli dispatch through the shared sandbox. This PR resolves the observed storm by making the existing backoff gate actually function.

### DIFF
--- a/apps/web/lib/routing/cli-pool-status.test.ts
+++ b/apps/web/lib/routing/cli-pool-status.test.ts
@@ -60,6 +60,13 @@ describe("parseRetryAfterSeconds", () => {
     expect(parseRetryAfterSeconds("try again in 10 seconds")).toBe(10);
   });
 
+  it("parses hour/minute reset windows (ChatGPT/Codex subscription limits)", () => {
+    expect(parseRetryAfterSeconds("try again in 5 minutes")).toBe(300);
+    expect(parseRetryAfterSeconds("rate limit reached. try again in 2 hours")).toBe(7200);
+    expect(parseRetryAfterSeconds("reset in 2h 30m")).toBe(9000);
+    expect(parseRetryAfterSeconds("you've hit your limit — retry in 1h")).toBe(3600);
+  });
+
   it("returns null for unparseable input", () => {
     expect(parseRetryAfterSeconds("rate limited")).toBeNull();
     expect(parseRetryAfterSeconds("429 Too Many Requests")).toBeNull();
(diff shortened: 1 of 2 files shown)


markdbodman:
### Follow-ups filed + kernel decision

This PR is the **reactive** layer (once a provider 429s, back off + fall back). Two follow-ups cover the proactive and structural layers:

- **BI-15068745** — proactive multi-provider load-spreading for Build Studio fan-out. Routed through the kernel (`principle_decide`, high confidence, composite 8.17 / margin 2.51, no commandment conflict): **reuse the existing capacity-weighted scorer** (`pipeline.ts:379` already discounts endpoints >80% utilization) rather than build a new load-balancer. Two gaps to close: seed approximate rate limits for CLI/subscription providers (their caps read 0% utilization today, so the penalty never fires), and tier-route fan-out phases instead of hard-pinning one provider.
- **BI-CE7DE053** — bound codex-cli concurrency through the shared sandbox (mirror `withLocalInferenceLock`).

The three compose: proactive spreading → reactive fallback (this PR) → hard concurrency bound.

PR #2786, opened 2026-07-11
fix(self-upgrade): COPY config/ into Docker build + guard against build-context divergence

markdbodman:
## Why the last self-upgrade failed

Self-upgrade run **SUR-BCFB72BB** (corrective **BI-062CFB41**) failed at the `docker-build` step:

```
Module not found: Can't resolve '../../../../config/seed-content-paths.json'
  ./apps/web/lib/integrate/seed-contribution-fit.ts:1:1
```

**Root cause.** PR #2776 added `config/seed-content-paths.json` at the **repo root** and statically imported it from `apps/web/lib/integrate/seed-contribution-fit.ts`. The production `Dockerfile` copies a **narrow allowlist** into the `next build` stage (`apps/web/`, `packages/`, `docs/professions/`, a few named root files) and **never copied `config/`**.

**Why CI didn't catch it.** The required *Production Build* gate runs `pnpm --filter web build` (plain `next build`) against the **full checkout**, where the JSON is on disk → green. The real Docker image is only built by `publish-image.yml` on `push: tags: v*` / manual dispatch — **never on PRs**. So the narrow-COPY Docker build is only exercised at self-upgrade / tag-release, and the divergence shipped green and detonated every upgrade to a target ≥ #2776.

This is a **recurring class** (cf. `docs/professions/registry.json`, `packages/db` data JSON, undici hoist — all previously bitten: BI-165756F8, BI-C0B9B7AA).

## Fix

1. **`COPY config/`** into the build stage (`next build`) and the init stage (`generate-tools-snapshot.js` loads the same `mcp-tools → contribution-review → seed-contribution-fit` graph). Mirrors the identical `docs/professions/` pattern already in both stages.
2. **`dockerfile-build-context.guard.test.ts`** — derives the build-stage COPY allowlist *from the Dockerfile* and fails the **Unit Tests** gate if any build-time static import in `apps/web` escapes into a repo dir the image never copies. Adding a `COPY <dir>/` line auto-satisfies it — no duplicated list to drift.

Approach chosen via `principle_decide` (2026-07-11) — kernel recommended the static guard (composite 1.52, high confidence, *Architecture Over Shortcuts*) over building the full image on every PR (CI cost / OOM pressure) or broadening COPY to the whole tree (image bloat).

## Verification

- Guard goes **red** on the unfixed tree (exactly 1 violation: `config/seed-content-paths.json`), **green** after the `COPY` (0 violations), across 3254 source files — all other escaping imports resolve under `packages/` / `docs/professions/` or are test-only.
- Docker `--target build` (reproduces the exact self-upgrade `next build`) run locally to confirm the module now resolves.

## Follow-up (not in this PR)

The build-failure classifier mislabeled this module-not-found as `bundle-boundary-static-import` because NFT warnings appeared above the fatal error in the log — so BI-062CFB41 points at the wrong playbook. Flagged separately.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -43,6 +43,15 @@ COPY scripts/set-hooks-path.mjs ./scripts/
 COPY apps/web/ ./apps/web/
 COPY packages/ ./packages/
 COPY docs/professions/ ./docs/professions/
+# Root-level config data statically imported at build time — e.g.
+# apps/web/lib/integrate/seed-contribution-fit.ts imports
+# ../../../../config/seed-content-paths.json. Without this COPY the Next.js build
+# fails "Module not found" inside the image even though plain `next build` (CI)
+# passes, because CI builds the full checkout while the image build only sees this
+# narrow allowlist. This detonated self-upgrade SUR-BCFB72BB (BI-062CFB41); the
+# dockerfile-build-context.guard.test.ts guard now fails CI if any build-time
+# import escapes into a repo dir this stage doesn't COPY. Mirrors docs/professions.
+COPY config/ ./config/
 COPY docker-entrypoint.sh ./
 RUN pnpm install --frozen-lockfile
 RUN pnpm --filter @dpf/db exec prisma generate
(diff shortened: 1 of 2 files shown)


markdbodman:
**In-image functional verification.** Ran `docker build --target build` locally (the exact stage that runs `next build` inside the image — the step that failed self-upgrade SUR-BCFB72BB). With this PR's `COPY config/`:

```
#33 [build 10/10] RUN NODE_OPTIONS="--max-old-space-size=8192" ... next build
#33 0.541   Creating an optimized production build ...
#33 56.27 ✓ Compiled successfully in 56s
#33 122.7 Route (app)
#33 DONE 123.2s
#34 exporting to image ... DONE
```

No `Module not found: config/seed-content-paths.json` — the import resolves inside the narrow-COPY build context. (CI's Production Build gate exercises plain `next build`, not the image, so this in-image run is the piece CI structurally can't cover — which is exactly why the guard test was added.)

PR #2968, opened 2026-07-14
fix(self-upgrade): local-changes ledger uses content diff, not a sha range (BI-75C4A412)

markdbodman:
## Problem

The local-changes ledger (Private/Public Change Segregation) lists the commits an operator has kept private — not shared upstream. It used a **two-dot commit range** `${remote}/${branch}..${installBranch}` in `git log`.

But shared changes flow upstream and **return via the self-upgrade merge under a rewritten SHA** (upstream squash-rebases contributions — origin/main is confirmed squash-rebased). The operator's original commit is therefore never reachable from upstream, so the two-dot range keeps counting *already-shared* changes as "private" — **over-reporting the private delta**.

## Fix

Compare by **content**, not SHA: `--cherry-pick --right-only` over a three-dot symmetric range (`${remote}/${branch}...${installBranch}`). `--cherry-pick` drops any install-branch commit that is patch-equivalent to one already upstream; `--right-only` keeps only the install-branch side. What remains is the genuinely-private delta.

## Verification

Functionally verified in a throwaway repo (make a change on the install branch, re-apply it upstream under a different SHA, plus one genuinely-private commit):
- **two-dot (old):** reports both the shared change and the private one ❌
- **three-dot `--cherry-pick --right-only` (new):** reports only the private one ✅

Unit test updated to assert the three-dot range + `--cherry-pick`/`--right-only` flags. Local scoped typecheck skipped (source-only worktree, no `node_modules`); CI runs the authoritative typecheck + tests.

Fixes `BI-75C4A412`. (Supersedes an earlier unmerged branch fix `e6513b2c4`.)

### DIFF
--- a/apps/web/lib/self-upgrade/local-changes-ledger.test.ts
+++ b/apps/web/lib/self-upgrade/local-changes-ledger.test.ts
@@ -39,8 +39,15 @@ describe("collectLocalChanges", () => {
     expect(res.available).toBe(true);
     expect(res.changes).toHaveLength(1);
     expect(res.changes[0].subject).toBe("feat: keep this private");
-    // log invoked with the upstream..installBranch range
-    expect(run).toHaveBeenLastCalledWith(expect.arrayContaining(["origin/main..dpf/install", `--format=${LEDGER_LOG_FORMAT}`]));
+    // BI-75C4A412: content diff, not sha range — three-dot symmetric range with
+    // --cherry-pick --right-only drops commits already upstreamed under a
+    // different (squash-rebased) sha.
+    const logArgs = run.mock.calls.at(-1)?.[0] as string[];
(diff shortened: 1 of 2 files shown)


markdbodman:
Superseding this: origin/main already resolves BI-75C4A412 with a stronger fix. Main uses a three-dot **tree diff** (`git diff --numstat --no-renames <remote>/<branch>...<install>`, per-file {path,added,deleted}), explicitly discarding commit identity because squash-merges destroy it. This PR used `git log --cherry-pick --right-only` (patch-id matching), which still over-reports squash-merged changes since a squash has a different patch-id than its constituent commits. Main's approach is correct; closing in favor of it. (This branch was based on a commit that has since been squash-rebased off main, hence the conflict.)

PR #3074, opened 2026-07-16
fix(sandbox): stream writeSandboxFile content via stdin, not argv, to stop secret leak (BI-AFEF038A)

markdbodman:
## What & why (BI-AFEF038A)

Follow-up to **BI-1291B677** (OAuth token leaked into BuildActivity via the CLI dispatch command surface). That fix was a source-level try/catch + redaction; this closes the underlying mechanism.

`writeSandboxFile` base64-encoded the file content and embedded it **directly in the `docker exec` command**:

```
docker exec <c> sh -c "echo '<base64>' | base64 -d > <path> && chmod <mode> <path>"
```

Because the payload lives in **argv**, any secret staged this way (the CLI OAuth token, the mcp-config JWT, codex `auth.json`) is visible in `ps` / `/proc`, container exec logs, and a rejected exec `Error` whose `.cmd`/`.message` carries the base64 — which decodes straight back to the secret.

## The fix

Stream the base64 payload via the child's **stdin** instead of argv, using `docker exec -i <c> sh -c "base64 -d > <path>"` — the exact shape already proven in `codex-cli-adapter.ts` (BI-2685C1E5).

**Before**
```ts
await sandboxExec()(
  `docker exec ${userFlag}${containerId} sh -c "${mkdir}echo '${b64}' | base64 -d > ${path}${chmod}"`,
);
```
**After**
```ts
await dockerExecWriteStdin({
  containerId,
  innerCommand: `${mkdir}base64 -d > '${path}'${chown}${chmod}`, // secret-free
  input: b64,                                                     // → child stdin
  asNodeUser,
});
```

- New shared `dockerExecWriteStdin` helper (spawn + stdin write + timeout/close loop). Its rejection carries `stderr` + exit code but **never** the input.
- The target `path` (and `mkdir`/`chown`/`chmod` operands) are now **single-quoted**, so a path can't break out of `sh -c` either.
- `cli-adapter.ts`'s five inline `echo '<b64>' | base64 -d` writes (prompt / system / **mcp-config JWT** / **OAuth token** / runner script) now route through `writeSandboxFile` (added an opt-in `chownNodeUser`). The per-caller try/catch + redaction become belt-and-suspenders rather than the primary control.

## Tests

- **New** `agent-cli-runtime.test.ts` proves content flows via stdin: asserts the raw secret **and its base64** are absent from argv (`expect(args.join(" ")).not.toContain(secret)` / `.not.toContain(b64)`), that stdin decodes back to the exact bytes, path single-quoting, `chown`/`mkdir`/`chmod` assembly, and secret-free rejection on non-zero exit / timeout.
- `cli-adapter.test.ts` write-inspection tests updated to read the mocked `writeSandboxFile` `(path, content, mode, chownNodeUser)` instead of decoding argv base64.

Verified locally: touched vitest files **34/34 pass**; `tsc --noEmit` on `@dpf/web` **clean** (8GB heap; the pre-commit hook OOMs at its default 4GB, so this commit used the hook-sanctioned `DPF_SKIP_TYPECHECK=1` — CI gates typecheck).

## Risk / edge cases

- **Large files / binary content**: stdin is unbounded (also dodges `ARG_MAX`/E2BIG); base64 round-trip preserves exact bytes — `base64 -d` ignores whitespace so the missing trailing newline vs the old `echo` is inert.
- **Path escaping**: paths are now single-quoted (were unquoted). All callers pass fixed `/tmp/...` / `/root/.codex/...` paths with no single quotes, so behavior is unchanged for real inputs while removing the breakout surface.
- **Out of scope**: other `echo '<b64>' | base64 -d` writers (`tool-script.ts`, `sandbox.ts`, `coding-agent.ts`, `sandbox-pack.ts`) are separate callers not named in this BI; the existing redaction in `mcp-tools.ts` / `build-evidence-pack.ts` still covers them.

Security change — please review; auto-merge intentionally **not** enabled.

Generated with Claude Code

### DIFF
--- a/apps/web/lib/integrate/sandbox/agent-cli-runtime.test.ts
+++ b/apps/web/lib/integrate/sandbox/agent-cli-runtime.test.ts
@@ -1,0 +1,256 @@
+// apps/web/lib/integrate/sandbox/agent-cli-runtime.test.ts
+//
+// Security coverage for BI-AFEF038A: writeSandboxFile must stream file content
+// (potentially a secret — OAuth token, mcp-config JWT, codex auth.json) through
+// the child's stdin, NEVER embedded in the `docker exec` command. If the content
+// were in argv it would show up in `ps` / /proc, container exec logs, and any
+// rejected-exec error string — which is exactly the leak this BI closes.
+
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { EventEmitter } from "events";
+
+// ── Spawn mock infrastructure (mirrors codex-cli-adapter.test.ts) ─────────────
+
+interface MockProcess extends EventEmitter {
+  stdin: {
+    write: ReturnType<typeof vi.fn>;
+    end: ReturnType<typeof vi.fn>;
+    on: ReturnType<typeof vi.fn>;
+  };
+  stdout: EventEmitter;
+  stderr: EventEmitter;
+  kill: ReturnType<typeof vi.fn>;
+  /** All bytes handed to stdin.write, concatenated. */
+  stdinData: string;
+}
+
+function makeMockProcess(): MockProcess {
+  const proc = new EventEmitter() as MockProcess;
+  proc.stdinData = "";
+  proc.stdin = {
+    write: vi.fn((data: unknown, cb?: (err?: Error | null) => void) => {
+      proc.stdinData += String(data);
+      cb?.();
+      return true;
+    }),
+    end: vi.fn(),
+    on: vi.fn(),
+  };
+  proc.stdout = new EventEmitter();
+  proc.stderr = new EventEmitter();
+  proc.kill = vi.fn();
+  return proc;
+}
+
+const mockSpawn = vi.fn();
+
+vi.mock("@/lib/shared/lazy-node", () => ({
+  lazyChildProcess: () => ({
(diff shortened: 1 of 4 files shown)


markdbodman:
**Security code review (Claude, orchestrating session): APPROVED pending functional verification.**

Reviewed `dockerExecWriteStdin` + `writeSandboxFile` and the test assertions. The secret now flows only over the child's stdin; `innerCommand` is secret-free (`base64 -d > '<path>'`), the target path and mkdir/chown operands are single-quoted (sh -c breakout closed), stdin EPIPE is handled, the promise settles once, and a rejected write carries stderr/exit but never the input. Tests prove the contract directly: argv contains neither the raw secret nor its base64 encoding, stdin decodes back to the exact secret bytes, and the error path is secret-free (34/34 pass). Mirrors the already-proven `writeContainerFileViaStdin` shape (BI-2685C1E5).

**Left non-auto-merge intentionally.** The unit tests mock `spawn`, so they prove the argv/stdin split but not that a real container's `base64 -d` reconstructs the file byte-for-byte from stdin on the live sandbox path (structural ≠ functional). Since `writeSandboxFile` is on the critical path of every CLI-adapter run, this wants one real sandbox write verified against a running container before merge — which requires a deploy (operator-gated). Holding for @markbodman's functional check.

PR #3079, opened 2026-07-16
fix(backlog): gate low-severity [reference-doc] proposals from auto-filing to the backlog (BI-18685188)

markdbodman:
## What & why

The architecture-review / doc scanner emits a large recurring batch (~15-20 per pass) of low-severity `[reference-doc] … would benefit from …` doc-polish improvement proposals, and **every one auto-files into the backlog** as a `BI-IMP-*` item with **no severity gate**. These are founder-reserved canonical-doc authorship (AGENTS.md, kernel principles, architecture specs) — Mark hand-defers ~35/day. It re-inflates the open backlog every scanner pass and is the dominant recurring noise source.

This ships the **minimal, reversible default**: withhold **only** `low`-severity `[reference-doc]`-titled proposals from auto-filing to the backlog, at **both** auto-file paths. `medium`/`high`/`critical` reference-doc proposals and **all** non-reference-doc proposals file **exactly as today**. **Nothing is lost** — gated proposals remain as `ImprovementProposal` rows (`backlogItemId` null) in the Improvements view for founder review; they are only withheld from becoming `BI-*` backlog items. Additive, reversible, and scoped to the doc-suggestion auto-file path only (not the PlatformIssueReport/log-signature pipeline).

## Precedent this mirrors

Skill proposals are **already** excluded from the backlog (`category: { not: "skill" }` — they run the governed skill-revision approve/rollback lifecycle instead; reconcile.ts guardrail *"Skill proposals stay separate"*). This extends that exact exclusion pattern conservatively to the low-severity reference-doc doc-polish class.

## Implementation

A single shared predicate `isLowSeverityReferenceDocProposal(proposal)` in `apps/web/lib/process-spine/reference-doc-promotion.ts` (the module that owns the `REFERENCE_DOC_RE = /^\[reference-doc\]/i` marker). It keys on the exact markers the scanner sets in `promoteReferenceDocFindings` (`title: "[reference-doc] …"`, `severity: "low"`). Both call sites use it, so the rule lives in one place:

- `apps/web/lib/evaluate/improvement-backlog-reconcile.ts` — the orphan-drain reconcile now filters the fetched proposals through the predicate before filing.
- `apps/web/lib/mcp/packs/contribution-hive-pack.ts` — `propose_improvement` → `ingestBacklogItem` skips the creation-time auto-file when the predicate matches (the proposal is still created; only the backlog projection is withheld).

## Tests

Exhaustive predicate + behavior coverage (29 tests pass; `tsc --noEmit` clean):
- low-severity reference-doc → **NOT filed** (both paths)
- medium/high/critical reference-doc → **filed** (both paths)
- low-severity NON-reference-doc (a normal proposal) → **still filed** (no over-suppression)
- skill → **still excluded as before** (reconcile DB filter contract unchanged)
- marker matched as a title prefix only, case-insensitive, tolerant of leading whitespace/empty severity

## Open policy questions for @markbodman (why this is NOT auto-merge)

Suppressing a class of auto-detected items is a founder-reserved product decision, so this is a **regular, non-auto-merge PR**. Two threshold questions for review:

1. **Which severities to keep?** This default gates `low` only and keeps `medium`/`high`/`critical`. If the desired floor is higher (e.g. keep `high` only), the predicate is the single place to change.
2. **Do reference-doc suggestions belong in the backlog at all** vs. a **separate doc-improvement queue**? This PR keeps them as `ImprovementProposal` rows in the Improvements view; a dedicated queue would be a larger follow-up.

Design-Grounding-Decision: BI-18685188 (source of truth); extends the skill-exclusion precedent in `docs/superpowers/specs/2026-06-06-work-intake-unification-design.md` §3.3. No new contract.

Generated with Claude Code


markdbodman:
**Code review (Claude, orchestrating session): APPROVED — awaiting your policy call on the threshold.**

Verified the predicate and both call sites. `isLowSeverityReferenceDocProposal` requires `severity === "low"` **and** a `[reference-doc]` title *prefix* (reuses the canonical `REFERENCE_DOC_RE`), so it's tightly scoped to exactly the scanner's noise class. Reconcile drops those orphans while the skill exclusion stays intact via the unchanged DB where-clause; the creation path wraps only the backlog projection in `if (!autoFileSuppressed)`, so the `ImprovementProposal` row is still created — nothing is lost, the suggestion just stays in the Improvements view. The fail-safe direction is right: empty/unknown severity → `false` → **files** (never silently suppresses on missing data), and a mid-title marker still files (prefix-only). 29 tests, touched files typecheck-clean.

**Non-auto-merge is correct.** This ships the minimal reversible default (gate `low` reference-doc only). The two questions the item reserves for you — which severities to keep, and backlog-filter vs. a separate doc-improvement queue — are genuine product-policy calls; this waits on your decision, not a deploy. The evidence that motivated the conservative default: you hand-deferred ~35 of these in a single day, and skill proposals were already excluded by the same precedent.

PR #3081, opened 2026-07-16
refactor(portal): prefer typed agent role metadata over keyword inference in hive-mind resolver (BI-REFACTOR-B6A61421)

markdbodman:
## What

Replaces the portal context hive-mind resolver's keyword role inference (`inferRole`) with typed agent role metadata as the primary source, keeping keyword sniffing only as a fallback for not-yet-populated agents.

## Why

`resolveHiveMindCandidates` inferred each agent's hive-mind role by keyword-sniffing its text and logged a one-time warning on that fallback (BI-REFACTOR-B6A61421). Role should be first-class, typed metadata authored at the source, not re-derived heuristically on every request.

## Changes

- **`packages/db/data/agent_registry.json` (authoring source of truth):** add a typed `role` (`builder|reviewer|architect|tester|operator|specialist`) to **all 67 seeded agents**, each derived from its existing definition (name + capability domain) — no invented roles. Distribution: specialist 28, operator 23, reviewer 8, architect 4, builder 3, tester 1.
- **`Agent` Prisma model:** add nullable `role String?`. Carried from the registry so the JSON stays the authoring home; the column exists only because the resolver reads the Prisma `Agent` table and no existing column carries a hive-role.
- **Migration `20260716140000_add_agent_hive_mind_role`:** hand-authored, additive, metadata-only `ADD COLUMN` (`@migration-safety: data-safe`), seed-backfilled.
- **`seed.ts`:** load `role` from the registry into the Agent row (fix-the-seed, not runtime).
- **`hive-mind-resolver.ts`:** select `role` and prefer it in `resolveAgentRole`; fall through to `inferRole` (with the retained one-time warning) only when the typed role is absent. `inferRole` removal is **DEFERRED** per BI-REFACTOR-B6A61421 until seed + live agent data is fully populated (code comment added).
- **Tests:** resolver prefers typed role even when keywords would infer a different one (no warn) vs. falls back + warns when absent; registry hygiene guard keeps every agent tagged with a canonical role token.

## Verification (local)

- `apps/web` resolver tests: 3 passed.
- `packages/db` registry hygiene tests: 4 passed.
- `tsc --noEmit` clean on both `apps/web` and `packages/db` (0 errors).

## Migration note

This PR includes a Prisma migration (additive `Agent.role` column). It is **left non-auto-merge** as prod-risky by convention; please merge via the queue after review.

Design-Grounding-Decision: BI-REFACTOR-B6A61421 — portal context overlay architectural review gap (2026-05-20); replace hive-mind keyword role inference with typed agent role metadata, deferring inferRole removal until population is confirmed.

Generated with Claude Code


markdbodman:
**Code review (Claude, orchestrating session): APPROVED pending merge + migration on deploy.**

Verified the full path. The migration is textbook data-safe — `ALTER TABLE "Agent" ADD COLUMN "role" TEXT` (nullable, no default → metadata-only, no table rewrite, no existing row can violate), with NULL transparently falling through to the retained keyword fallback. The resolver adds `role: true` to the Prisma select and `resolveAgentRole` prefers `normalizeRole(agent.role)` when present (no warn, no sniffing), only calling `warnOnceForInferredRole` + `inferRole(text)` when the typed role is absent. `inferRole` is correctly retained with a DEFERRED-removal comment citing the BI — its removal stays gated on full population confirmation, exactly as scoped. Roles are authored in `agent_registry.json` (source of truth) and backfilled by the seed for all 67 agents; a registry hygiene guard (mirroring the value-stream guard) enforces a role on every agent from canonical tokens. Resolver + hygiene tests pass; tsc clean on apps/web + packages/db.

Good substrate catch: the `resolveAgentRole`/`normalizeRole` scaffolding already existed on main but was dead (no column, no select, no seed) so every agent always hit the fallback — this wires it end-to-end.

**Non-auto-merge is correct** — it carries a Prisma migration (applies via promote.sh on the next self-upgrade), so it should merge under your eye. The change is backward-compatible either way (unpopulated agents keep working via the fallback).

PR #3099, opened 2026-07-16
feat(hygiene): schedule the runtime-artifact janitor in observe/dry-run mode (BI-DBF3F426)

markdbodman:
## What & why

BI-DBF3F426's residual work: the Docker-side janitor (`scripts/runtime-artifact-janitor.mjs`, which reaps orphaned per-branch CI build images + stray/foreign compose projects) was **built but never scheduled** — it only ran as a manual CLI. This gives it a scheduled home, mirroring `runtime-target-janitor.ts`.

## OBSERVE-ONLY — this schedule logs, it never deletes

This reaps Docker containers/images, so `destructive-actions-require-explicit-go` governs. The scheduled function runs the janitor's **detection in DRY-RUN mode only**: it **logs what it WOULD reap** (structured: what / why-classified-orphan / age) and records a summary. **It deletes nothing.** Turning on real `--apply` reaping is deliberately **deferred to a separate, founder-gated operator action** — not part of this PR.

### Where dry-run is enforced (no `--apply` reachable from the schedule)

`apps/web/lib/queue/functions/runtime-artifact-janitor.ts`, three layers:

1. The only argv the schedule ever passes the CLI is `OBSERVE_SCAN_ARGS = Object.freeze(["--json"])` — no `--apply`, no `--live`. The CLI defaults to dry-run anyway.
2. `assertObserveArgs()` **throws at module load** if `OBSERVE_SCAN_ARGS` ever grows an apply/live token (regression tripwire) — called at import and again at the spawn call site.
3. The runner re-checks the CLI's own reported `mode` and **refuses** any scan whose `mode !== "dry-run"`.

The reaping functions (`reapImage` / `reapComposeProject`) live only inside the CLI's `main()`, reachable only when it parses `--apply`/`--live`, which this file never emits.

Further gated by **`DPF_RUNTIME_ARTIFACT_JANITOR_ENABLED` (default OFF)** so even the harmless dry-run scan is opt-in. Daily at 05:20. When docker/git aren't reachable (e.g. inside the portal container), detection degrades to "unavailable" — logs, never throws.

### Orphan-detection criteria the founder is asked to review before enabling `--apply`

Quoted from `scripts/lib/runtime-artifact-janitor.mjs`:

- **Build image** — a candidate once it matches `^dpf-local-integration-[a-z0-9-]+-build$` **and** is idle `>= staleness-days` (default **7**). Rationale in the lib: *"per-branch CI build images are prohibited substrate — runtime gates must go through the shared `local-integration-ci` lease. So *any* image matching the tag convention is a reap candidate once it crosses the staleness threshold."*
- **Compose project** — reaped only when ALL hold: (1) it is **not** the root `dpf` project; (2) it has **no live worktree** backing it (a `dpf-<topic>` project whose worktree is still present is KEEP regardless of age); (3) its newest container is older than staleness-days. Foreign (non-`dpf`) idle projects are also candidates.

Volumes are explicitly **out of scope** (`down` without `--volumes`, per `never-wipe-db`).

## Tests

`runtime-artifact-janitor.test.ts` asserts: `OBSERVE_SCAN_ARGS` is exactly `["--json"]` with no reaping flag; `assertObserveArgs` throws on `--apply`/`--live`; disabled-by-default skip (scan not even invoked); the dry-run path returns the correct would-reap summary (only REAP verdicts counted, KEEP excluded); non-dry-run mode is refused; and unavailable-detection degrades to a skip.

## Verification note

The web toolchain (vitest/tsc) is env-blocked in this source-only worktree (pnpm store not materialized — confirmed at root, and the known-good sibling `contributor-inventory-sync.ts` throws the same module-resolution errors under an ad-hoc tsc run). CI's Typecheck + Unit gates are authoritative. Code mirrors the proven `runtime-target-janitor.ts` / `contributor-inventory-sync.ts` patterns.

**Not auto-merge** — activating any janitor near the Docker lifecycle wants founder review, and `--apply` activation is explicitly deferred to founder sign-off.

Design-Grounding-Decision: BI-DBF3F426

Generated with Claude Code


markdbodman:
**Code review (Claude, orchestrating session): APPROVED as observe-only — safe to merge; --apply activation stays yours.**

Verified the safety-critical property directly: there is no reachable path from this schedule to the janitor's destructive `--apply`/`--live`. Confirmed in `runtime-artifact-janitor.ts`: `OBSERVE_SCAN_ARGS = Object.freeze(["--json"])` is the only argv handed to the CLI; `assertObserveArgs()` runs at module load AND at the spawn call site and throws if an apply/live token ever creeps in (a real regression tripwire); the spawn passes only the frozen observe args; and a defense-in-depth check refuses any scan whose CLI-reported `mode !== "dry-run"`. The reaping functions live only in the CLI's `main()` behind `--apply`/`--live`, which this file never emits. Activation is further gated behind `DPF_RUNTIME_ARTIFACT_JANITOR_ENABLED` (default OFF) plus the scheduled-functions flag. Mirrors the proven `runtime-target-janitor.ts` pattern. Tests assert the would-reap accounting (KEEP excluded), that `assertObserveArgs(["--apply"])` throws, that non-dry-run is refused, and that the disabled flag skips the scan entirely.

This satisfies the `destructive-actions-require-explicit-go` commandment: it observes and logs what it *would* reap, deletes nothing, and turning on real reaping is a separate founder-gated action. Non-auto-merge is correct — it's near Docker lifecycle, so it merges under your eye; the orphan-detection criteria (image tag `^dpf-local-integration-[a-z0-9-]+-build$` + staleness ≥7d; compose project not-root + no live worktree + stale) are quoted in the PR body for your review before you ever flip `--apply` on.

PR #3105, opened 2026-07-16
fix(discovery): make the (from,to,type) tuple the canonical InventoryRelationship key to end cross-run P2002 (BI-PIR-7d69a445)

markdbodman:
## Why this is safe — the evidence (decision: TUPLE-CANONICAL)

`relationshipKey` on `InventoryRelationship` is **source-scoped provenance**, not the row's global identity. The evidence that the `(fromEntityId, toEntityId, relationshipType)` tuple is the real identity:

- **Nothing looks up `InventoryRelationship` by `relationshipKey`.** A repo-wide grep found no `findUnique`/`findFirst`/`where` on `InventoryRelationship.relationshipKey` outside `discovery-sync.ts` (the upsert conflict target + the stale sweep, both re-pointed here). So its `@unique` bought nothing except the crash.
- **Per-run `relationshipKey` provenance is preserved** on `DiscoveredRelationship` via `@@unique([discoveryRunId, relationshipKey])`. Every source key still gets its own provenance row linked to the shared canonical `InventoryRelationship`. Demoting `InventoryRelationship.relationshipKey` to non-unique loses no information.
- **The tuple is already the persisted-entity identity** — `@@unique([fromEntityId, toEntityId, relationshipType])` — and it is left unchanged.
- **No existing data violates the tuple uniqueness.** Verified on the live estate: `0` duplicate-tuple groups (expected, since the tuple `@@unique` was already enforced). Dropping the `relationshipKey` unique index is a metadata-only loosening that cannot fail on existing rows.

## What was wrong (the cross-run P2002)

The upsert targeted `where: { relationshipKey }` but its `create` wrote the tuple. When a **prior run** had already persisted the same real relationship under a **different** `relationshipKey`, the new run missed on `relationshipKey`, took the **create** path, and violated the tuple `@@unique` → `P2002` aborted the whole persistence `$transaction`.

Before / after conflict target:

```diff
- where: { relationshipKey: relationship.relationshipKey },
+ where: {
+   fromEntityId_toEntityId_relationshipType: {
+     fromEntityId,
+     toEntityId,
+     relationshipType: relationship.relationshipType,
+   },
+ },
```

Now a prior run's tuple re-appears → the upsert **UPDATES** it (refreshing `lastSeenAt` / `lastConfirmedRun` / `relationshipKey`) instead of trying to create a duplicate.

## This completes part 2, atop #3054's part 1

- **Part 1 (#3054, merged):** in-run tuple dedup — a `Map<tupleKey, persistedId>` so two keys that collapse to one tuple **within a run** upsert once. Kept intact.
- **Part 2 (this PR):** the cross-run/concurrent case — the tuple was persisted by a **prior** run under a different `relationshipKey`. Re-points the upsert conflict target at the tuple `@@unique`, drops the now-wrong global uniqueness on `relationshipKey`, and re-keys the created-vs-updated accounting and the stale sweep off the tuple (`existingRelationshipByTuple`; stale rows marked by `id`).

## Changes

- **Migration** `20260716230000_inventoryrelationship_tuple_canonical`: `DROP INDEX "InventoryRelationship_relationshipKey_key"`. `@migration-safety: data-safe` — metadata-only, loosening a constraint, cannot fail on existing data; the canonical tuple `@@unique` is unchanged.
- **Schema**: `relationshipKey String @unique` → `relationshipKey String`. The steward-reviewed removal is recorded in `schema-regression-guard.mjs` `INTENTIONAL_FIELD_REMOVALS` (the column is kept; only its uniqueness is dropped).
- **`discovery-sync.ts`**: tuple-targeted upsert; tuple-keyed created/updated accounting; tuple-keyed stale sweep (marks stale rows by `id`); `update` block refreshes `relationshipKey` to the latest observed key.
- **Tests**: new part-2 cross-run regression test (below); part-1 in-run dedup test and the scope/source-attribution suites adjusted to the tuple conflict target and id-based sweep. All 10 discovery-sync tests pass locally.

## Regression test assertion

A tuple persisted by a **prior** run under `relationshipKey` **K1**, re-presented on a **new** run under a **different** `relationshipKey` **K2**, is **UPDATED** — exactly one upsert whose `where` is the tuple compound unique, counted as `updatedRelationships: 1` / `createdRelationships: 0`, **not swept stale**, and no `P2002` — while `DiscoveredRelationship` still records the new run's own **K2** provenance linked to the one shared canonical row.

## Verification

- `vitest run` on `discovery-sync.test.ts` + `discovery-sync-scope.test.ts` + `discovery-sync-source-attribution.test.ts` → **10 passed**.
- `prisma generate` against the modified schema confirms the compound-unique input `fromEntityId_toEntityId_relationshipType`; `tsc --noEmit` on the change is clean (the only tsc output is pre-existing `@dpf/storefront-templates` workspace-link noise, unrelated to this change).

Design-Grounding-Decision: BI-PIR-7d69a445

Generated with Claude Code

### DIFF
--- a/packages/db/prisma/migrations/20260716230000_inventoryrelationship_tuple_canonical/migration.sql
+++ b/packages/db/prisma/migrations/20260716230000_inventoryrelationship_tuple_canonical/migration.sql
@@ -1,0 +1,31 @@
+-- BI-PIR-7d69a445 (part 2): make the (fromEntityId, toEntityId, relationshipType)
+-- tuple the CANONICAL identity of an InventoryRelationship and demote
+-- relationshipKey to non-unique.
+--
+-- InventoryRelationship carried TWO uniques: relationshipKey (source-scoped) and
+-- the compound (fromEntityId, toEntityId, relationshipType) (persisted-entity
+-- level). Entity dedup can map two distinct discovered refs onto ONE persisted
+-- entity id, so relationships with DISTINCT relationshipKeys resolve to the SAME
+-- tuple. The discovery-sync upsert targeted `where: { relationshipKey }` but its
+-- `create` wrote the tuple, so when a PRIOR run had already persisted that tuple
+-- under a different relationshipKey the create path violated the compound
+-- @@unique -> P2002 aborted the whole persistence transaction. Re-pointing the
+-- upsert at the tuple (in the same commit) fixes that; this migration removes the
+-- now-wrong global uniqueness on relationshipKey.
+--
+-- relationshipKey stays as a plain column (still written for provenance /
+-- back-compat). Nothing looks up InventoryRelationship by relationshipKey
+-- (no findUnique/where on it outside discovery-sync), and per-run relationshipKey
+-- provenance is preserved on DiscoveredRelationship via its own
+-- @@unique([discoveryRunId, relationshipKey]) -- so demoting this @unique loses
+-- nothing.
+--
+-- @migration-safety: data-safe: dropping a UNIQUE index is a metadata-only
+-- catalog change (no table rewrite, no heap scan of existing rows) and only
+-- LOOSENS a constraint, so it cannot fail on existing data. No new/tightened
+-- constraint is introduced. The canonical tuple @@unique
+-- ("InventoryRelationship_fromEntityId_toEntityId_relationshipType_key") is
+-- UNCHANGED and remains enforced, so no duplicate tuple can appear as a result of
+-- this change (verified: 0 duplicate-tuple groups on the live estate before
+-- migration).
+DROP INDEX "InventoryRelationship_relationshipKey_key";
--- a/packages/db/prisma/schema.prisma
+++ b/packages/db/prisma/schema.prisma
@@ -4530,7 +4530,15 @@ model InventoryEntity {
 
 model InventoryRelationship {
   id                      String                   @id @default(cuid())
-  relationshipKey         String                   @unique
+  // BI-PIR-7d69a445 (part 2): relationshipKey is SOURCE-SCOPED provenance that
+  // varies per source/scope for the same real relationship, so it is NOT the
+  // canonical identity — it is intentionally non-unique. The canonical identity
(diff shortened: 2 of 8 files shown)


markdbodman:
**Code review (Claude, orchestrating session): APPROVED — implements the evidence-backed canonical-key decision; your confirm is the only gate.**

This completes BI-PIR-7d69a445 part 2 atop #3054's part 1, and it matches the evidence exactly. Verified:
- **Migration** drops `InventoryRelationship_relationshipKey_key` — a pure metadata-only loosening (drops a unique index), `@migration-safety: data-safe`; cannot fail on existing data.
- **Upsert re-pointed to the tuple**: the conflict target is now `fromEntityId_toEntityId_relationshipType` (confirmed against the generated client type), and the created-vs-updated accounting + stale sweep now key off `existingRelationshipByTuple` (by row id), not `relationshipKey`. So a prior run's tuple re-presented under a different `relationshipKey` UPDATEs the one canonical row instead of throwing P2002.
- **Safety evidence is airtight**: no other code does a unique lookup on `InventoryRelationship.relationshipKey` (only discovery-sync, re-pointed here); per-run provenance is preserved on `DiscoveredRelationship.@@unique([discoveryRunId, relationshipKey])`; and the agent queried the live DB — **0 duplicate-tuple groups**, so nothing was masked by the dropped constraint.
- **Regression test** proves the cross-run case: prior run K1 + new run K2 on the same tuple → single UPDATE, no P2002, DiscoveredRelationship still records K2. Part-1 in-run dedup stays green (10/10).
- `relationshipKey` is added to the schema-regression-guard `INTENTIONAL_FIELD_REMOVALS` allowlist (needed since that guard runs in CI too).

**The decision embedded here — the `(from,to,type)` tuple is canonical, `relationshipKey` is source-scoped provenance — is yours to confirm.** The code makes it; the evidence backs it; you own the data model. Non-auto-merge is correct: it carries a schema migration that applies on the next self-upgrade, so it should merge under your eye.

PR #3107, opened 2026-07-16
docs(integrations): per-family connector benchmark scorecards (BI-INT connector epic)

markdbodman:
## What

Per-family, module-by-module **connector benchmark scorecards** for all 10 `BI-INT-*` connector families. This is the next level of detail below the merged #3098 family-level coverage matrix: where #3098 answered "per family, how much does DPF have?", this doc answers "per family, **which specific capability** is built / partial / missing, with the file that proves it, and the cheapest next slice to build."

Deliverable: `docs/superpowers/plans/2026-07-16-connector-family-benchmark-scorecards.md` (links the #3098 matrix doc).

## For each family
- A built / partial / missing capability table, each row citing a **code path as evidence** (`INTEGRATION_COVERAGE_MATRIX`, native adapters under `apps/web/lib/integrate/<vendor>/`, `NATIVE_INTEGRATIONS`, `SUPPORTED_INTEGRATIONS`, estate-discovery collectors, `integration-shared`).
- A one-paragraph **target-parity definition** (what "benchmarked-complete" means concretely).
- The **minimal first connector slice** (cheapest module that moves coverage), so the build is spec'd for whenever credentials + sequencing arrive.

## Honesty posture (why it's valuable)
- This is the **benchmark artifact (analysis)**. Actually building connectors is separate, needs **live provider credentials** to functionally verify, and needs the founder's **A-vs-B sequencing** (#3098 §4). Stated at the top.
- Every DPF-coverage claim is grounded in cited code; unresolvable cells say **"unknown from code"**; unverifiable Make module lists say **"baseline unverified (Make external)"** — which is **all ten families** (module counts are external to the repo).
- #3098's **7 open unknowns are carried forward**, not silently resolved.
- Two grounded corrections to #3098's substrate counts: the in-code matrix is **13 native + 9 benchmark = 22 rows** (not 14+8), and estate discovery ships **8 collectors** (not "unifi + 1").

## Refs
Families: BI-INT-6C83D1 (identity), BI-INT-E76A95 (CRM), BI-INT-8D4F72 (comms), BI-INT-F23BC6 (HR), BI-INT-A5B9E3 (finance), BI-INT-C1D472 (service desk), BI-INT-1AB7D8 (project/work), BI-INT-2CE9F1 (docs), BI-INT-3DF4A2 (device/RMM), BI-INT-47B8C3 (MSP). Builds on #3098.

Docs-only → satisfies the Spec/Plan/Doc gate by construction.

Generated with Claude Code

PR #3111, opened 2026-07-16
feat(integrations): register the built WhatsApp Business + Instagram Business adapters in the native catalog (BI-INT-8D4F72)

markdbodman:
## What

Registers the **WhatsApp Business** and **Instagram Business** native integration adapters in `NATIVE_INTEGRATIONS` so they surface as available-to-connect integrations. Advances the Communications connector family (**BI-INT-8D4F72**).

## Why

Both adapters were **built and fully tested** (client + connect-action + preview + 3 passing test files each, ~700+ LOC, landed in #3011) — identical in shape to the already-registered `hubspot`/`facebook-pages` adapters — but were **never added to the catalog**, so they did not appear as available integrations. Their dedicated routes already exist:

- Page: `app/(shell)/platform/tools/integrations/{whatsapp,instagram}-business/page.tsx`
- API connect: `app/api/integrations/{whatsapp,instagram}-business/connect/route.ts` (calls the adapter's connect-action)

Route resolution is **convention/directory-based and already wired** — the only gap was the catalog descriptor. #3107's per-family connector benchmark scorecards flagged this divergence. No TODO/FIXME/feature-flag/"do not register" marker exists in either adapter; this was an oversight, not an intentional hold.

## Changes

- `native-integration-catalog.ts`: add `whatsapp-business` + `instagram-business` to the `NativeIntegrationId` union and two `NATIVE_INTEGRATIONS` descriptors. Values derived from each adapter's own code — `provider: "facebook"` and `integrationId` matching the adapter `INTEGRATION_ID`; `enables` reflect the adapters' actual read-first probes (WhatsApp phone/template readiness + localized template coverage; Instagram profile/media/comment context).
- `integration-coverage-matrix.ts`: add matching coverage rows (required by the coverage-matrix test that maps every native integration to a coverage row).
- Tests: assert both new ids surface through the catalog and connection-catalog; update the native-count assertions (13 → 15 native) in `connection-catalog.test.ts` and `tool-marketplace-readiness.test.ts`.

## Scope / verification

- Registration makes the integrations **available-to-connect only**. The operator supplies their **own Meta credentials** through the existing connect-action.
- **Functional verification against live WhatsApp/Instagram (actually connecting to Meta) is a separate, credentials-gated step and is out of scope here** — it needs the user's Meta access token / WABA / IG business account.
- Touched suites: `npx vitest run` → **33 passed / 11 files** (both adapters' client/connect-action/preview, catalog, coverage-matrix, connection-catalog, tool-marketplace-readiness, integrations page). Touched files typecheck clean (`tsc --noEmit`). Module-size gate passes.

Design-Grounding-Decision: BI-INT-8D4F72

Generated with Claude Code


markdbodman:
**Review (Claude, orchestrating session): APPROVED — clean finish of genuinely-built work.**

Verified the guardrails you ran: both adapters have the full client + connect-action + preview + test shape identical to the registered HubSpot adapter, the connect-actions implement real probe→credential.upsert flows, git history shows they landed in #3011 (a docs PR) as a registration oversight rather than an intentional hold, and no TODO/flag/"not ready" markers exist. Confirmed they weren't already registered elsewhere (the `channel-types.ts` whatsapp hit is a different concept — an employee-reachability medium). Descriptor values are derived from each adapter's own PROVIDER/INTEGRATION_ID/probe capabilities, not invented. Route wiring is convention-based and already present (`app/api/integrations/<id>/connect/route.ts`), so the descriptor was sufficient. Good catch that `integration-coverage-matrix.ts` needed coverage rows or its parity test would fail — that's the same class of drift-guard we hit on the janitor cron.

Auto-merge is appropriate: complete, freshly-green adapters (33 tests), no functional/behavioral risk from registration alone — it makes them available-to-connect, and the user supplies their own Meta auth via the connect-action. Live-Meta functional verification stays a separate credentials-gated step, correctly noted as out of scope.

PR #3116, opened 2026-07-16
refactor(wip): derive the enforced WIP cap from active WorkCapsules across surfaces, not BS builds alone (BI-937128F6)

markdbodman:
## What & why (BI-937128F6)

Residual wiring slice for **unifying the enforced WIP cap** so it derives from active WorkCapsules + shared-resource pressure across ALL delivery surfaces — not Build Studio builds alone. The classification substrate (`apps/web/lib/build/unified-wip.ts`: `summarizeUnifiedWip`, `gatingPool`, `contendsOnBsSandbox`) was already built; this wires it into the enforced gate.

### Before → after
- **Before:** both build-entry gates (`createFeatureBuild`, `promote_to_build_studio`) counted **only Build Studio builds** (`prisma.featureBuild.count(...)`) and compared against a hardcoded `BUILD_WIP_CAP = 3` (BS-only).
- **After:** both gates derive admission from the **unified, pool-aware** model. A new unit of work contends on a finite resource pool (`unified-wip.gatingPool`) and is blocked **only when that pool is saturated across all surfaces' active WIP**. New: `apps/web/lib/build/unified-wip-query.ts` (`loadActiveUnifiedWip` reads FeatureBuilds + non-BS WorkCapsules + active nonprod leases) and `wip-cap.ts` `decideUnifiedWip` / `assertUnifiedWipCapacity`.

### Derivation implemented
- The gate for a promote / new BS build asks: **is the `bs-sandbox` pool saturated?** Pressure = active WIP contending on that pool across surfaces; capacity = `WIP_POOL_CAPACITY["bs-sandbox"]`.
- **The live `bs-sandbox` limit stays `BUILD_WIP_CAP = 3` — today's effective limit is preserved exactly.** Only build-studio work contends on the sandbox (`contendsOnBsSandbox`), so the population of that pool is exactly the active BS builds and the gate is behavior-for-behavior identical to before.
- **bs-sandbox pressure is read from the authoritative `FeatureBuild` count, NOT from build-studio capsules.** A build-studio `WorkCapsule` is created in `"working"` status and its status is **not** transitioned when the build reaches a terminal phase (`capture-build-pr.ts` stamps the PR but leaves it `"working"`), so counting those capsules would over-count finished builds and wrongly **tighten** the cap. Every *other* surface's active WorkCapsule IS counted (for the non-sandbox pools + the cross-surface picture).

### Capacity constants (named + documented in `wip-cap.ts` `WIP_POOL_CAPACITY`)
| pool | capacity | source | enforced by this cap? |
|---|---|---|---|
| `bs-sandbox` | `BUILD_WIP_CAP` = **3** | existing constant — **unchanged** | ✅ yes (the two build-entry gates) |
| `shared-lease` | `SHARED_LEASE_WIP_CAP` = **1** ⚠️ **chosen — please confirm** | documents the singleton `:3001` `NonProductionEnvironmentLease` arity (`@@unique(activeKey)`) | ❌ enforcement already lives in the lease workflow; this cap does not re-enforce it |
| `host-worktree` | unbounded | external work in its own worktree, ungated today (AGENTS.md §17) | ❌ intentionally ungated |
| `none` | unbounded | human / git-webhook, not resource-bound | ❌ |

### ⚠️ Founder decision flagged
`SHARED_LEASE_WIP_CAP = 1` is a **chosen** number (grounded in the singleton lease's real arity, not an arbitrary formula). It is **not** enforced by this cap at the two build-entry call sites — those gate only `bs-sandbox`. The constant is defined for the unified model / future wiring. Please confirm the value (and whether the shared-lease pool should ever be enforced here rather than solely by the lease workflow).

## Verification
- Unit tests are the bar (green locally, 5 affected suites / 69 tests + wider sweep 12 suites / 259 tests):
  - `decideUnifiedWip` / `assertUnifiedWipCapacity` per pool; `WIP_POOL_CAPACITY` pins `bs-sandbox === BUILD_WIP_CAP === 3`.
  - **Regression pin:** `decideUnifiedWip("bs-sandbox", n).admitted === !wipCapReached(n, 3)` for every n; end-to-end pin that a new BS build is blocked at 3 active builds and admitted at 2 **even with many external capsules in flight** (they contend on other pools, must not affect the sandbox gate).
  - `loadActiveUnifiedWip` sources bs-sandbox from `FeatureBuild.count`, excludes build-studio capsules, classifies shared-lease by active `:3001` lease match.
- `tsc --noEmit` on the touched files is clean (the only 2 remaining project errors are pre-existing schema-drift in this source-only worktree's junctioned Prisma client — e.g. `@dpf/db/healthcare-*` — unrelated to this change).

**Live functional verification is deferred to founder post-merge review** — this changes a live build-concurrency gate, so the cap correctly gating real concurrent builds across surfaces should be confirmed on the running install. **Not** auto-merge on purpose.

## Design grounding
Extends existing substrate (`unified-wip.ts` + EP-UNIFIED-TRACKING doctrine, AGENTS.md §17); no new contract. Source of truth: `unified-wip.ts` header + BI-937128F6. Doc updated in the same PR (`docs/architecture/agent-client-capability-parity.md`).

Design-Grounding-Decision: BI-937128F6

Generated with Claude Code


markdbodman:
**Code review (Claude, orchestrating session): APPROVED — careful unification that preserves the live gate; your confirm on one constant.**

Verified the correctness that matters most: the enforced bs-sandbox limit is **unchanged** — `WIP_POOL_CAPACITY["bs-sandbox"] = BUILD_WIP_CAP` (3), and `assertWipCapacity`/`wipCapReached`/`BUILD_WIP_CAP`/`BuildWipCapError` are all kept intact for their existing consumers (`dispatch-bet.ts`, `evaluate-drain.ts`). So this unifies the *source* of the count (now pool-aware across all surfaces' active WIP), not the limit. The regression pin — `decideUnifiedWip("bs-sandbox", n).admitted === !wipCapReached(n, 3)` for every n, plus the end-to-end assertion that a new BS build still blocks at 3 even with 10 external capsules in flight — is exactly the right guard that other pools don't leak into the sandbox gate.

The best part is the substrate catch: sourcing bs-sandbox pressure from the authoritative `FeatureBuild` count rather than build-studio WorkCapsules, because those capsules are created `"working"` and never transitioned to terminal (`capture-build-pr.ts` stamps the PR but leaves the status), so counting them would over-count finished builds and wrongly tighten the cap. That's a real bug avoided, documented in code. 259 tests pass; touched files typecheck clean.

**One thing is genuinely yours:** `SHARED_LEASE_WIP_CAP = 1` — the single chosen capacity, grounded in the singleton `:3001` lease arity but flagged for you (it isn't enforced at the build-entry gates today; the lease workflow owns it). Confirm the value and whether this cap should ever enforce the shared-lease pool. Non-auto-merge is correct: it changes a live build-concurrency gate, so it merges under your eye, and live verification across concurrent real builds is deferred to your post-merge check.

PR #3129, opened 2026-07-17
refactor(platform-config): extract pure platform-dev-config helpers into a domain-core module (BI-OPT-FAT-ACTIONS — platform-dev-config slice)

markdbodman:
## Pure MOVE — behavior preserved

Behavior-preserving relocation of the side-effect-free helpers out of `apps/web/lib/actions/platform-dev-config.ts` into a new domain-core module `apps/web/lib/platform-config/platform-dev-config-core.ts`. Same pattern as the EA slice (#3118, merged), CRM slice (#3125), and Build slice (#3126). Helper **bodies relocated verbatim** — no logic/signature/contract changes. Orchestration (auth, prisma, `revalidatePath`, `fetch`, fs/exec via lazy-node, the `"use server"` wrappers) stays in the action file and imports the helpers back.

### Helpers extracted (by theme, one module)
- **GitHub-token validation primitives:** `detectAuthMethod`, `parseScopesHeader`, `scopeSatisfies`, `parseExpiryHeader` — plus the `ValidateTokenInput` / `ValidateTokenResult` shapes, **re-exported** from the action module (`export type { ... }`) so existing public importers keep resolving them unchanged.
- **Git-merge string/error primitives:** `shellQuote`, `gitErrorText`, `isMissingGitReferenceError`.

### LOC
- `apps/web/lib/actions/platform-dev-config.ts`: **1331 → 1257** (shrinks below its `scripts/module-size-baseline.txt` entry — passes the ratchet; module-size check green).
- New `apps/web/lib/platform-config/platform-dev-config-core.ts`: **114 LOC** (< 800 soft ceiling).

### Existing tests green, unchanged
The three behavior-pin suites (`platform-dev-config.validate.test.ts`, `.apply.test.ts`, `.reencrypt.test.ts` — 25 tests) pass **unchanged**. Added `platform-dev-config-core.test.ts` (19 focused assertions) for the extracted helpers.

### Zero unused imported-back symbols — how verified
[...]

PR #3118, opened 2026-07-16
refactor(ea): extract pure EA helpers into a domain-core module (BI-OPT-FAT-ACTIONS — EA slice)

markdbodman:
## What

Pure **MOVE** refactor: extract the side-effect-free logic out of the fat `apps/web/lib/actions/ea.ts` action file into two new EA domain modules, mirroring the already-merged Slice-A tax-remittance extraction (`af179f845` → `lib/finance/tax-remittance-core.ts` + `tax-remittance-validation.ts`). This is the **EA slice** of `BI-OPT-FAT-ACTIONS` under `EP-PLATFORM-OPTIMIZATION` ("reduce the bodies behind the consolidated boundaries").

## Extracted helpers

**`apps/web/lib/ea/ea-core.ts`** — pure structured-view mutation logic:
- `sortStructuredMutationRecords`
- `resequenceStructuredChildren`
- `insertStructuredChild`
- `deriveStructuredWarnings`
- types `StructuredMutationRecord`, `StructuredConformanceWarning`

**`apps/web/lib/ea/ea-validation.ts`** — pure reference-status guards:
- constants `REFERENCE_COVERAGE_STATUSES`, `REFERENCE_PROPOSAL_STATUSES`
- types `ReferenceCoverageStatus`, `ReferenceProposalStatus`
- assertions `assertCoverageStatus`, `assertProposalStatus`

The thin orchestration — `auth()` checks, all prisma reads/writes, the `$transaction` blocks, `revalidatePath`, the fire-and-forget Neo4j sync, and the `"use server"` action wrappers — stays in `ea.ts`, which now imports the extracted helpers back.

## Behavior preservation

**Byte-for-byte preserved.** Identical function bodies (verbatim relocation), identical error messages, unchanged call sites, and no change to any server-action external contract. Callers of the `ea.ts` server actions are unaffected.

- `apps/web/lib/actions/ea.ts`: **1,055 → 917 LOC** (module-size ratchet shrinks; baseline was 1055).
- Existing `apps/web/lib/actions/ea.test.ts` is the **behavior pin** — it passes **unchanged** (25 tests, zero edits; the actions delegate to the moved helpers).
- New direct unit tests for the now-pure helpers: `ea-core.test.ts` (12 tests) + `ea-validation.test.ts` (4 tests) = 16 new tests. Full run: **41 passed** across the 3 files.

[...]

PR #3133, opened 2026-07-17
refactor(compliance): extract pure compliance helpers into a domain-core module (BI-OPT-FAT-ACTIONS — compliance slice)

markdbodman:
## What

Pure **MOVE** of the side-effect-free logic out of `apps/web/lib/actions/compliance.ts` into a new domain-core module `apps/web/lib/compliance/compliance-core.ts`. Behavior is byte-for-byte preserved — this mirrors the merged EA slice (#3118), the Build slice (`build-actions-core.ts` / `businessBriefJsonPayload`), and the platform-dev-config slice (#3129).

The relocated helpers are the pure Prisma **data-payload builders** (each `*Input` → create/update field map: trim/normalize/`?? null` defaults + ID generation) and the dashboard `regulationSummaries` **projection**. Orchestration — `auth()`, `prisma`, `$transaction`, `revalidatePath`, the calendar/audit-log side effects, and the `"use server"` wrappers — stays in the action file and imports the helpers back. The prisma-typed JSON casts (`Regulation.applicability → Prisma.InputJsonValue`) stay at the call sites, so the core module carries **no `@dpf/db` value/type coupling**.

## LOC

| file | before | after |
| --- | --- | --- |
| `apps/web/lib/actions/compliance.ts` | 1067 | **866** |
| `apps/web/lib/compliance/compliance-core.ts` | — | 365 (new, < 800 cap) |

Baseline is 1068; the shrink passes the module-size ratchet.

## Helpers extracted (20)

Create/update field-map builders for: regulation, obligation, control, risk-assessment, incident, corrective-action, audit, audit-finding, submission (create+update pairs), plus `evidenceCreateFields` (create-only; **reused** by both `createEvidence` and the `supersedeEvidence` transaction, which build the identical row), and `mapRegulationSummaries` (dashboard projection).

Left as orchestration on purpose: the `onboardRegulation` and `supersedeRegulation` transaction bodies (loops / `buildNextRegulationVersion`-derived rows with different defaults), and `verifyCorrectiveAction` (inline `new Date()`).

## Behavior preservation

- Existing behavior-pin test `apps/web/lib/actions/compliance.test.ts` (16 tests) is **UNCHANGED** and green — the pin.
- Added `apps/web/lib/compliance/compliance-core.test.ts` — 28 new unit tests asserting each builder's trim/default/employeeId-fallback/conditional-inclusion behavior, ID prefixes, and the projection shape.
- `tsc --noEmit` on `apps/web`: **zero errors in touched files**; the only remaining project errors are the pre-existing stale-junctioned `@dpf/db/healthcare-*` subpath artifacts (unrelated to this change).

## Zero unused imports

After extraction, every imported-back symbol was word-boundary grepped in the action file and confirmed to occur ≥1 time **outside** the import block (create/update builders each appear twice = import + call site; `evidenceCreateFields` three times). The now-unneeded ID generators (`generateAssessmentId`, `generateIncidentId`, `generateActionId`, `generateAuditId`, `generateFindingId`, `generateEvidenceId`, `generateSubmissionId`) were **dropped** from the `@/lib/compliance-types` import (they now live inside the core builders); `generateRegulationId`/`generateObligationId`/`generateControlId` are retained (still used by the onboarding/supersede transactions). `node scripts/check-no-type-reexport-in-use-server.mjs` passes — no type is re-exported through the `"use server"` module.

Generated with Claude Code

### DIFF
--- a/apps/web/lib/actions/compliance.ts
+++ b/apps/web/lib/actions/compliance.ts
@@ -18,14 +18,24 @@ import {
 // @/lib/actions/compliance-helpers directly.
 import {
   generateRegulationId, generateObligationId, generateControlId,
-  generateAssessmentId, generateIncidentId, generateActionId,
-  generateAuditId, generateFindingId, generateEvidenceId, generateSubmissionId,
   validateRegulationInput, validateObligationInput, validateControlInput,
   type RegulationInput, type ObligationInput, type ControlInput,
   type RiskAssessmentInput, type IncidentInput, type CorrectiveActionInput,
   type AuditInput, type FindingInput, type EvidenceInput, type SubmissionInput,
   type OnboardingInput,
 } from "@/lib/compliance-types";
+import {
+  regulationCreateFields, regulationUpdateFields,
+  obligationCreateFields, obligationUpdateFields,
+  controlCreateFields, controlUpdateFields,
+  riskAssessmentCreateFields, riskAssessmentUpdateFields,
+  incidentCreateFields, incidentUpdateFields,
+  correctiveActionCreateFields, correctiveActionUpdateFields,
+  auditCreateFields, auditUpdateFields,
+  auditFindingCreateFields, auditFindingUpdateFields,
+  evidenceCreateFields, submissionCreateFields, submissionUpdateFields,
+  mapRegulationSummaries,
+} from "@/lib/compliance/compliance-core";
 
 // ─── Regulation ─────────────────────────────────────────────────────────────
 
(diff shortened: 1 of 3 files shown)


markdbodman:
**Code review (Claude, orchestrating session): APPROVED — the cleanest, biggest-shrink slice, and it dodged the #3129 trap.**

Verified: `compliance.test.ts` (16 tests) is untouched (not in the diff) — the behavior pin holds against the refactored action file. compliance.ts shrinks 201 lines (1067→866, the largest reduction of any slice); the 20 extracted field-map builders + `mapRegulationSummaries` projection are independently tested (28 assertions). Zero unused imports — the 7 now-internal ID generators were dropped from the import, the 3 still used by the onboarding/supersede transactions retained, all remaining counts ≥2. And critically: **zero exports added to the "use server" file** — I re-confirmed no `export type {`/`export { type` re-export, so the `check-no-type-reexport-in-use-server` guard passes and there's no repeat of #3129's Production Build break.

Good judgment on the deliberate holds: `onboardRegulation`/`supersedeRegulation` transaction rows use *different* defaults (`category ?? "other"` vs `?? null`) so they're genuinely not the same field map — extracting them would have caused subtle behavior drift, and leaving `evidenceCreateFields` as one helper reused by both create + supersede (byte-identical shape) is exactly right. Keeping the Prisma JSON casts at the call sites keeps the core module free of `@dpf/db` coupling.

Non-auto-merge is fine for your eye; behavior risk is near-zero (unchanged pin + pure move + guard-verified).

PR #3463, opened 2026-07-23
[HOLD - blocked on BI-EA221325] arm the route budget ratchet — frozen baseline

markdbodman:
The final step of EP-UX-SYSTEM Phase 1: flips the wall-of-text gate from **reporting** to **enforcing**. "Every iteration adds more wall of text" is now a failed required check.

## What arms

- `route-budget-baseline.json` ships **`bootstrapped: true`**, **200 routes**, each with the full axis set — words, controls, fields, choices, sub-legible controls, **buriedPrimaryAction** (the reachability axis from #3457), axe violations, and the normalized structural snapshot.
- Measured by the `workflow_dispatch` update path (#3458) on a branch off current main, so it reflects the real portal in the gate's own environment.

From here a changed route may not exceed its frozen baseline on any axis, and its accessibility structure may not change shape. Net-new routes meet the absolute shell budgets. Re-baseline deliberately via the dispatch path as surfaces migrate.

## Snapshot redaction, completed

Freezing the real baseline surfaced two things the live DOM does that raw ARIA snapshots can't store:

1. **Volatile values** (deploy SHAs, ids, timestamps, counts) — redacted in #3462, so a deploy doesn't read as a structure change.
2. **Secret-shaped labels** — an integration-setup screen renders a "paste your key" field labelled with a PEM header, captured in the accessible-name text. A UI label, not a secret, but it trips secret scanning and shouldn't be persisted verbatim. `normaliseSnapshot` now redacts PEM blocks/headers and JWTs to `<pem>` / `<jwt>` before the value patterns, on both freeze and compare sides, idempotently.

**Verified:** gitleaks reports no leaks on the committed baseline — no `-----BEGIN` survives in any stored snapshot. 23 ratchet tests pass, including the PEM/JWT redaction case (inputs assembled from fragments so this file carries no contiguous secret-shaped literal).

## The league table this records

The founder's concern, quantified — the worst-first migration order for Phase 2:

| words on arrival | route |
|---|---|
| ~15,000 | /platform/audit/authority |
| 5,258 | /platform/ai/skills |
| 4,272 | /inventory (a cockpit) |

Design-Grounding-Decision: arms and refines the gate in docs/superpowers/specs/2026-07-22-holistic-ux-system-and-agent-codification-design.md section 6 L4 / section 7.1. Implementation in apps/web/lib/ux-budget/ratchet.ts and the data file apps/web/lib/ux-budget/route-budget-baseline.json.
Docs-Impact-Decision: no doc change — snapshot redaction is internal to the ratchet; the arming procedure is already documented in docs/platform-usability-standards.md (UX Budgets).
Module-Size-Decision: no baselined file affected; ratchet.ts well under 800 LOC.
Process-Spine-Decision: refines an existing merged mechanism; no spec/plan change.


markdbodman:
**Holding as draft — do not merge.**

The enforcing sweep on this PR flagged **98 of 200 routes** with "accessibility tree changed shape" plus small word drift (268→270, 287→289). This PR changes only `ratchet.ts` and the baseline data — nothing that alters any route's rendered content — so these are **false regressions**: the sweep is not reproducible across runs.

Cause: every run provisions an ephemeral Postgres and re-seeds a fresh fixture org, so wall-clock-derived and generation-ordered content differs run to run. The existing normalisation removes SHAs/ids/timestamps but not relative times, ordering, or small counts.

Arming in this state would fail unrelated PRs on ~half of all routes and get the gate disabled — the exact failure this epic exists to prevent. Filed as **BI-EA221325** with candidate fixes (structure-only ARIA projection is the likely one; names are where the non-determinism lives, and the gate's purpose is hierarchy shape, not label text).

The sweep stays in reporting mode on main, which is honest — it measures 200/226 routes and produces the §7.1 league table. Re-freeze and flip `bootstrapped:true` once two consecutive runs on the same commit agree.

PR #3484, opened 2026-07-24
feat(kernel): slug-rename pass + first corpus migration to data-architect (BI-5FE47130)

markdbodman:
EP-DECISION-TIER-REBALANCE **step 3 — unblocked**. Steps 1 (#3478) and 2 (#3481) are merged.

Seed behaviour changes, deliberately: `seedWikiKernel` gains a rename pass ahead of the upsert. It is a no-op until a migration entry exists, and the list ships empty.

## The spec asked for something the substrate couldn't do

Migration must "preserve decision history". It couldn't.

A principle's slug is derived from its path — `principles/<name>` in the kernel, `professions/<profession>/<name>` in a corpus — and `upsertWikiPage` keys on `(organizationId, slug)`. So moving a principle **changes its slug**: the seed finds no row for the new slug, creates a fresh one, and strands the original.

`WikiPage` has no `previousSlug`/`aliasOf` column and `WikiPageVersion` cascades on delete. So a `git mv` forces a choice between:

- **orphaning** the old row — a duplicate principle plus a stale Qdrant point, exactly the class [#3482](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/3482) found live (five hits for a principle the DB holds one of); or
- **deleting** it, which destroys its version history.

## Fix: rename in place, before the upsert

`WIKI_SLUG_MIGRATIONS` is an append-only `{from, to, reason}` list applied by `seedWikiKernel` ahead of seeding. The existing row is found under its new key and updated in place — **the row id survives**, and with it version history, decision references, and the existing Qdrant point. **No schema change, no new column.**

Properties, each pinned by test:

| Property | Why it matters |
|---|---|
| **Idempotent** | Second run is a no-op; a fresh install seeds at the new slug and reports `already-migrated`. |
| **Overlay-safe** | Renames across all orgs — renaming only the kernel row would strand an org overlay at the old slug, re-creating the orphan. |
| **Refuses to merge** | If both slugs exist it logs and does nothing. Merging would pick a survivor and discard the other's history — the exact loss this prevents — and would breach `@@unique([organizationId, slug])`. |
| **No chains, no collisions** | Registry tests reject two sources mapping to one destination, and `a→b, b→c` chains that would make the pass order-dependent across installs. |

## I was wrong about needing a schema change

My first read was that this required a Prisma migration for a continuity column. Checking falsified it, and also killed the simpler-looking option:

- **Explicit `slug:` frontmatter** preserving the kernel slug *does* work mechanically — the seeder honours `frontmatter.slug ?? deriveSlug(...)`. **Rejected**: the `professions/` prefix is load-bearing for WSID scoping, so the page would move on disk and stay invisible to profession retrieval.
- **A `previousSlug` column** works, but carries information the rename makes unnecessary.

## The migration list ships empty, on purpose

The first cohort lands with the corpus move itself, so the rename and the file move are reviewed together rather than machinery arriving speculatively ahead of any caller.

## What the plan records so it isn't re-derived

The cohort is **53** `route-domain-specific` principles — not the spec's 39 — across ~7 professions, **11 of them commandments**. And **23 of 53 have no `sources`**, which WSID requires to publish. That's a signal rather than just friction: an uncited principle is often *platform-native doctrine*, not professional practice — `structural-verification-is-not-functional` is a commandment, and migrating it out would be wrong. Hence three buckets: migrate / keep-and-re-archetype / author-provenance-first.

**Next cohort:** `data-model` → `data-architect` (16 principles, zero commandments, 10 sourced).

## Verification

9 new tests; **1508 pass** across `packages/db/src`; `packages/db` typechecks clean; doc-index and doc-link gates pass. Local `apps/web` typecheck remains unreliable in this worktree (its `node_modules/@dpf/*` symlinks into a stale sibling worktree) — CI is authoritative.

Seed-Fit-Decision: install-local-only


markdbodman:
**Second commit added: the first cohort actually migrates.**

Nine single-context `data-model` principles move to the data-architect corpus, each with a `WIKI_SLUG_MIGRATIONS` entry so existing installs rename the row in place — id, version history, decision references and Qdrant point all preserved. Git records all nine as renames, so file history follows too.

**Deliberately excluded:**
- `db-fallback-explicit` — tagged both `mcp` and `data-model`. Moving a multi-context principle to one profession would silently remove it from the other's reach.
- The six uncited ones. WSID won't publish a profession page without a source, and three of them (`bundled-services-active-by-default`, `no-provider-pinning`, `zero-click-provider-setup`) look mis-tagged as `data-model` anyway — they're provider/platform doctrine. They need the bucket triage, not a move.

**Reference sweep.** Moving a file breaks every link to it. This repoints AGENTS.md (7 references, including the canonical §11 identity and schema-audit rules), 15 specs/plans/audits, and a hardcoded path in `seed-decision-perspective.ts` that would otherwise have pointed at a deleted file. Doc index regenerated, link integrity passes.

Kernel principle count: **100 → 91**.

Verification: 1508 pass in `packages/db/src`, 1095 across `apps/web` decision/wiki/mcp suites, module-size + doc gates clean.

PR #3522, opened 2026-07-24
docs(plan): deliberation engine content-generation build plan (BI-7B6B3C5C)

markdbodman:
Implementation plan for BI-7B6B3C5C — the root cause behind the health-verdict thread (BI-12E24186 / #3516 shipped the symptom containment).

## What
Phased plan to make plan-phase deliberations actually generate branch content:
1. **Thread artifact content** to branches (orchestrator drops it as `_content` today) — no behavior change.
2. **Call the model** (`routeAndCall`), capture the recommendation, feed the synthesizer — behind `DELIBERATION_LIVE_INFERENCE`, default OFF.
3. **Budget + per-build backoff** — hard prerequisite before enabling, because enabling flips on live paid inference on every plan gate.

## Why a plan, not a build
Enabling live inference is a cost/WWWD decision that's the operator's to make. This PR is doc-only and cost-neutral; it sequences the build so the engineering is ready when that go is given.

Refs: BI-7B6B3C5C

### DIFF
--- a/docs/superpowers/plans/2026-07-24-deliberation-engine-content-generation.md
+++ b/docs/superpowers/plans/2026-07-24-deliberation-engine-content-generation.md
@@ -1,0 +1,48 @@
+# Deliberation engine — branch content generation (BI-7B6B3C5C)
+
+**BI:** BI-7B6B3C5C — "Plan-phase deliberation engine never generates branch content → 100% insufficient-evidence → build/self-upgrade runaway."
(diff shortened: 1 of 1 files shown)


markdbodman:
Note: this PR also re-syncs `scripts/instruction-plane-baseline.txt` for AGENTS.md (112515→112550). That is a **pre-existing drift on main** — main's AGENTS.md was already over its committed baseline (a merge-union drop from a concurrent PR), which fails the Instruction Plane Guard on every PR based on current main. This PR does not modify AGENTS.md; `--update` just syncs the baseline to already-merged reality.

PR #3720, opened 2026-07-29
fix(inference): trust an explicit operator served-context pin above the estimated ceiling (BI-F4D3B9E9)

markdbodman:
## What

On a fully-local install, a restricted-sensitivity coworker turn (e.g. the COO with its 34-tool surface) has exactly one eligible endpoint: the bundled local model. Routing demands `estimatedInput × 1.5` of context (~30k for that surface), but `resolveServedContextTarget` clamped the operator's served-context pin to the resource-aware **estimate** — which caps a 24 GB / qwen3-coder-30B host at the 24,576 floor. Result: the pin silently no-oped, the context filter excluded the only cleared endpoint, and every tool-bearing confidential/restricted turn dead-ended in "No AI model can handle this request" (BI-F4D3B9E9; the live outage behind tonight's COO failure, verified via router excludedTrace + `[turn]` telemetry).

## Why the estimate is wrong to override an explicit pin

The ceiling is weights + KV arithmetic calibrated on dense models. qwen3-coder is MoE (GQA, 4 KV heads) — measured on the exact host the estimate capped: **DMR serves 40,960 tokens on the RTX 4090 and answers inference, VRAM 23.9/24.5 GB.** The module's own contract names the operator pin as "the mechanism for going higher."

## Fix

- An explicit `PlatformConfig local.servedContextTokens` pin is clamped only to `[build floor, MAX_LOCAL_CONTEXT_TOKENS]`; exceeding the estimated ceiling logs a loud warning and is honored (llama.cpp degrades by offloading; the reconcile's deferred/unreachable statuses catch a refused load).
- The **no-pin default keeps the conservative resource-aware math** — nothing changes for installs without an operator pin.

## Verification

- `vitest run lib/inference/local-model-context-reconcile.test.ts lib/inference/local-model-policy.test.ts`: 49/49 green, including the rewritten over-ceiling-pin case and a new measured-realistic 40k pin case.
- Live install: pin set to 40,960; DMR reloaded and answering; reconcile repairs `ModelProfile.maxContextTokens` on its cycle, restoring COO routing.

Remaining scope tracked in BI-F4D3B9E9: provision the serving default at install, per-turn tool-surface budgeting (`surfaceZone=overload`), and persisting the RouteDecisionLog trace on the no-eligible-endpoints throw path.

Local-CI-Override: worktree pnpm .bin cannot execute the pregate runner; targeted suites green as above and PR CI re-validates the full surface.

Seed-Fit-Decision: not-applicable — runtime policy module + tests; no seeded content changed.

### DIFF
--- a/apps/web/lib/inference/local-model-context-reconcile.test.ts
+++ b/apps/web/lib/inference/local-model-context-reconcile.test.ts
@@ -47,12 +47,22 @@ describe("resolveServedContextTarget — resource-aware (BI-3E614946)", () => {
     expect(await resolveServedContextTarget()).toBe(122_880);
   });
 
-  it("clamps an over-ambitious operator override DOWN to the resource-aware ceiling (no over-commit)", async () => {
+  it("trusts an explicit operator pin ABOVE the estimated ceiling, clamped only to MAX (BI-F4D3B9E9)", async () => {
     mockPrisma.platformConfig.findUnique.mockImplementation(
       configByKey({ host_profile: HOST_4090, "local.servedContextTokens": 200_000 }),
     );
-    // 200k pinned, but a 24 GB card on the 30B can only serve ~24,576 — clamp down.
-    expect(await resolveServedContextTarget()).toBe(RECOMMENDED_BUILD_CONTEXT_TOKENS);
+    // The weights+KV estimate says a 24 GB card on the 30B caps at the floor, but
+    // the estimate measurably under-sizes MoE models (40k serves fine on this
+    // exact host). An explicit pin is the operator's documented mechanism for
+    // going higher — honor it, bounded only by the hard MAX.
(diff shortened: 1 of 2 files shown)


markdbodman:
Calibration update from live verification: 40,960 (the value in the PR body) spilled model weights to CPU on the 24 GB 4090 — 18k-token prefill blew the 120 s call timeout. Final operator pin on this install: **32,768** — 23,405-token prefill in 3.3 s, fully GPU-resident, and the COO answers end-to-end on `provider=local model=qwen3-coder` with `executedTools=3`. The fix in this PR is unchanged (it's what makes the pin effective); the measured MoE KV economics are recorded in BI-F4D3B9E9 item (e) for the estimator recalibration.

PR #3886, opened 2026-08-01
fix(healthcare): stop the care-intake grant tests expiring on the calendar

markdbodman:
Every grant fixture in `care-intake-api-repository.test.ts` hard-coded an expiry of `2026-08-01T15:00:00Z` — **18 occurrences of the same literal**. That moment passed on 2026-08-01, and `issueCareIntakeResumeGrant` rejects a non-future expiry, so the suite began failing for **everyone on a date rather than on a change**.

It is currently red on main and blocking unrelated PRs — I hit it on a change that touches nothing in `lib/healthcare`, which is exactly why this class is worse than an ordinary flake: **the failure looks like the PR under review caused it.** Same shape as the wall-clock UX-sweep flake (BI-0C6C2153).

## Fix

Anchored to `Date.now() + 1h` via a single `futureExpiry()` helper, so the fixture keeps meaning *"a valid, unexpired grant"* whenever the suite runs.

All 18 sites used the **identical** literal, so none was deliberately testing an *expired* grant — the helper preserves intent rather than changing it. A test that genuinely needs an expired grant should say so with an explicit past date, and the comment says so.

**Verification:** 11/11 green in that file (was 1 failed / 10 passed).

Docs-Impact-Decision: test-only fixture repair, no route or user-visible behaviour changes.

Process-Spine-Decision: one-line-class test repair unblocking main; no capability or contract change, so no plan or spec is warranted.

No canonical seed content changed; recorded as global-default because the test fixture is platform substrate on every install.

Local-CI-Override: Test-only fixture repair unblocking main. 11/11 green in the affected file; no runtime, schema, or contract surface. GitHub CI runs the full suite here.

### DIFF
--- a/apps/web/lib/healthcare/care-intake-api-repository.test.ts
+++ b/apps/web/lib/healthcare/care-intake-api-repository.test.ts
@@ -89,6 +89,16 @@ function database() {
   };
 }
 
+// WALL-CLOCK TIME BOMB, fixed. Every grant fixture here hard-coded an expiry of
+// 2026-08-01T15:00:00Z. That moment passed on 2026-08-01 and `issueCareIntakeResumeGrant`
+// rejects a non-future expiry, so the suite began failing for everyone on a date rather
+// than on a change — the same class as the wall-clock UX-sweep flake (BI-0C6C2153).
+// Anchoring to `now` keeps the fixture meaning "a valid, unexpired grant" whenever the
+// suite runs. Any future test that needs an EXPIRED grant should say so explicitly with a
+// past date, not rely on the calendar catching up.
+const futureExpiry = () => new Date(Date.now() + 60 * 60 * 1000);
+
+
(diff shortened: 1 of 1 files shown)


markdbodman:
Closing as duplicate. #3883 and #3885 both fix the same wall-clock expiry in `apps/web/lib/healthcare/care-intake-api-repository.test.ts` and are already CLEAN in the merge queue ahead of this one. No need to spend further CI on a third identical fix.

For anyone tracing the failure: the single failing test is `issueCareIntakeResumeGrant > issues the raw token once only after an allow decision` — the suite pinned a literal expiry of `2026-08-01T15:00:00.000Z`, which went stale on the calendar today. It is the sole red across 622 files / 5387 tests, and it currently blocks unrelated PRs (e.g. #3881) until one of the fixes lands on main.

PR #3895, opened 2026-08-01
fix(discovery): defer the 381-card estate grid and repoint quality-issue owners

markdbodman:
Two independent fixes, both closing out recommendations from the Estate Discovery investigation.

## 1. The page was 34 screens tall — and the quality panel was only 12% of it

Measured on the live install: **31,006px**. Per block:

| block | height | share |
|---|---|---|
| **Attributed Estate — 381 product cards, uncapped** | 20,572px | **66%** |
| Bootstrap Discovery run + connections | 9,134px | 29% |
| Portfolio Quality panel (already fixed in #3807) | 3,849px | 12% |
| Operational Inventory | 3,184px | 10% |

The panel fix landed and works — 397 open issues now render as 42 cards. But it was never the problem. **Two thirds of the page is one card per DigitalProduct, rendered flat.**

That grid now previews 12 with a `Show all 381` toggle.

**Why deferring is right here, not just smaller.** `docs/platform-usability-standards.md` names progressive disclosure as the *sanctioned* fix for a surface over budget ("rewarded, never taxed" — measurement excises collapsed disclosure). And the page's own header copy states its purpose: resolve evidence quality here, then *"manage the owned estate from portfolio and product pages"*. The full roster belongs behind a click on this surface. Nothing is lost — the toggle reaches every card, each still links to its product estate view, and the total stays visible.

**Also removed 381 copies of one sentence.** Every card carried an identical 15-word note; at 381 cards that alone was ~5,700 words of duplicated DOM copy. Now stated once above the grid. Source-level prose-lint cannot see that win — it counts source occurrences, not rendered ones — which is why the new file still books a +21 textMass entry while the page gets dramatically lighter.

Arbitrary `text-[9px]`/`text-[10px]` sizes carried over from the old markup were migrated to the generated `text-dpf-caption` scale rather than copied, so the new component adds no token drift.

## 2. Quality-issue owners named an agent that does not exist (BI-01630E24)

All 9 coworker-owned types declared `owner: "coworker:estate-specialist"`. **No agent has that id** — it is in neither `COWORKER_SLUG_TO_CANONICAL_AGENT_ID` nor the live `Agent` table. It was written against the *display name* of `inventory-specialist`, which is "Digital Product Estate Specialist".

Every drift report therefore named an owner nothing could route to. Combined with no scheduled task covering these queues, that is a large part of why `lifecycle_unverified` + `catalog_match_ambiguous` grew from 107/120 to 215/220 over three months untouched.

Repointed to `coworker:inventory-specialist` — the slug the rest of the platform already uses (`DISCOVERY_TRIAGE_AGENT_ID`, agent-model-defaults, coworker-grant-consistency, workforce-seed, and the live `discovery-taxonomy-gap-triage-daily` task).

**Guard:** a new registry test asserts every `coworker:<slug>` owner resolves in the canonical map. **Canaried** — reverting one owner fails it. A non-existent owner is now a test failure, not a silent routing dead end.

**Two existing tests had codified the bug.** Both `quality-issue-registry.test.ts` and `quality-issue-drift-sweep.test.ts` asserted the drift owner *was* `coworker:estate-specialist` — locking in the broken value rather than catching it. Both corrected.

**Corrected from my earlier reading:** the two `Agent` rows (`AGT-WS-INVENTORY` named "inventory-specialist", and `inventory-specialist` named "Digital Product Estate Specialist") are an **intentional dual-seed pattern** documented in `agent-identity.ts` (BI-74FD6420) — slug rows stay first-class because FK consumers key on them, and the roster collapses the pair via `dropDualSeedAliasAgents`. Not a defect; nothing to fix there.

**This does not by itself make the queues get worked.** No scheduled task covers these two types yet. It removes the structural blocker so routing *can* land — see BI-0B420A1D, reopened for its unbuilt phase-2 auto-processing loop.

## Verification

- 6 new component tests; 44 pass across `components/inventory` and the discovery route.
- `packages/db` suite 1805 pass. The 2 failures in `inventory-entity-canonical-read-completeness` are **pre-existing load-dependent timeouts** — that file passes in isolation in 9.06s against a 5s per-test limit, and fails identically on an unmodified tree.
- `apps/web` and both `packages/db` typecheck configs clean.
- 31/31 preflight guards clean. Module-size, style/token drift and prose-lint all green.
- prose-lint baseline: **one hand-added entry** for the new file, no `--update` snapshot.

**NOT YET OBSERVED ON THE LIVE INSTALL** — needs a deploy. Expected page height drops from ~31,000px to roughly a third; the remaining bulk is the Bootstrap Discovery block (9,134px) and Operational Inventory (3,184px), untouched here.

## Local-CI attestation

Two sandbox-gate attempts reached **no verdict on this code**: one crashed at the first preflight guard before evaluating anything (exit `4294967295`, no lease claimed); one passed **31/31 guards** then queued at position 5 for the gate's full patience window. Neither was a guard or test failure.

The queue is saturated — capacity 1, and the currently-active run waited 22 minutes for admission behind 5 sessions. Both a crash and a cancel leave the claimKey (`local-ci:<session>:<sha>`) terminal, so each retry needs a fresh commit amend while adding another competitor to a single slot.

Local-CI-Override: Shared local-CI queue saturated (capacity 1; active run waited 22min; 5 sessions queued). Two attempts reached no verdict on this code — one crashed pre-evaluation, one passed 31/31 guards then exhausted its queue patience. Local evidence complete: 1805 db tests, 44 web tests, all typechecks clean, canaried owner guard. Operator chose this same override for the identical blocker on PR #3873; CI is authoritative.

Refs: BI-01630E24, BI-0B420A1D


markdbodman:
## Merge blocked by infrastructure, not by this change

**28 checks pass. 3 fail — all at the same step, for the same reason.**

`Secrets Scan`, `Analyze (go)` and `Analyze (python)` each fail at **"Upload SARIF to GitHub Code Scanning"** with `API rate limit exceeded for installation`.

The scans themselves succeed. Verified per-job:
- gitleaks logs `INF no leaks found` (5 commits scanned), *then* fails on upload.
- CodeQL logs `Exported results to SARIF (101ms)`, *then* fails on upload.

**Please do not read the red `Secrets Scan` as "probably fine".** It is fine *here*, but only because that specific log line says so — check it per-job rather than assuming this failure mode.

### What I tried
- Waited out a rate-limit window and re-ran: the workflow is not retryable (`This workflow run cannot be retried`).
- Rebased onto current main for a genuinely fresh run (also clearing 18 commits of base drift): **failed identically**.

Three fresh runs, same failure. Rebasing does not move it.

### Why it blocks
The PR reports `mergeStateStatus: UNSTABLE` / `mergeable: MERGEABLE`, which reads as "non-required checks failing, still mergeable". That reading is wrong for this repo — the **merge queue is stricter than the PR-level required set**. This PR reached queue position 2 and flipped to `UNMERGEABLE` while the PR itself still said `MERGEABLE`. I initially mis-called this as non-blocking; correcting that here.

### Not specific to this branch
PR #3854 was observed `UNMERGEABLE` in the same queue, which had grown to 18 entries. Filed as **BI-2701F2CE** — the org-wide installation API budget appears exhausted, which stalls merge throughput repo-wide.

### This branch's own state
Substantively green: Production Build, all test suites, Policy Guards, and the UX Route Sweep — which measured the change at **4,534 → 708 words (−84%)** on `/platform/tools/discovery` and **4,513 → 687 (−85%)** on `/inventory`.

Ready to merge the moment the SARIF uploads can complete.

PR #4006, opened 2026-08-05
fix(local-ci): stop the shared gate queue from wedging (BI-2C7F51BA)

markdbodman:
Closes BI-2C7F51BA.

Three compounding defects turned any interrupted local-CI gate into a persistent wedge of the shared admission queue — for every session on the host, not just the one that died. They compound: a killed gate leaks a record (1), the record can become unreapable (2), and a concurrent drain can stall admission indefinitely (3). Fixing one still leaves the queue able to wedge.

## 1 — the reaper never ran on the success path

`gate-worktree.mjs` registers a queue observer and releases only its **own**. `releaseDeadLocalQueueObserversForGate` was wired solely into `pregate.mjs`'s interrupted/revival recovery paths — which sit behind `pregate.mjs:88`'s `could not resolve worktree path` throw, so they are skipped on exactly the failure the cleanup exists for. The observed log was literally `queued gate revival skipped (could not resolve worktree path)`.

Nothing swept the directory unconditionally. 192 records had accumulated, 185 with dead pids, the oldest from 2026-07-29; a pregate sat at "position 3" for ~30 minutes and jumped to position 1 the moment they were swept by hand.

Gates now sweep on **startup**, immediately before registering, so every launch self-heals.

The sweep passes **no branch/sha/ownerSessionId filter, deliberately** — and says so in the code. A record leaked by a session that has since exited will otherwise never be reaped by anyone, because the only process that would scope a sweep to it is gone. Narrowing this to the current gate silently restores the leak.

## 2 — pid liveness was unsound (permanent wedge)

Liveness was `process.kill(pid, 0)`. Windows recycles pids aggressively, so a record whose gate died reads alive **forever** once the OS hands its pid to an unrelated process — a permanent queue-slot leak with no start time, no heartbeat, no TTL to bound it.

`(pid, startTime)` is unique where pid alone is not. Records now carry `processStartedAtMs`, and liveness compares it against the host's real process table (`Win32_Process` via PowerShell on Windows, `ps -Ao pid=,etimes=` elsewhere; memoized 15s because the reaper also runs inside the claim-retry loop). Two guarantees:

- **Verified live** — pid present with a matching start time. Never reaped, however long the gate runs. The TTL deliberately does not apply here.
- **Unverifiable** — no start time (pre-fix record) or the table could not be read. Falls back to pid-only liveness **plus** a 4h TTL, so nothing outlives a plausible gate run regardless of pid state.

A pid whose start time does not match is proven reuse: dead, reaped now. An unreadable process table returns `null`, not an empty map, so it can never be mistaken for "nothing is running".

## 3 — a waiter counted as itself as activity

`quiescence.ts` computed the soft `request.recent-tool-execution` blocker from a recency query over `ToolExecution`. Every MCP call writes a row **including read-only ones**, so a gate polling `get_quiescence_status` while waiting for the drain kept the blocker armed at count 1 — its own poll. The diagnostic used to inspect the stall did the same. The signal was measuring observation, not work.

The query now also excludes `auditClass = metrics_only` — the class `deriveAuditClassForTool` assigns a tool with neither a side effect nor external access. This extends the mechanism the `EDGE_AGENT_PREFIX` clause already established, and reuses the argument BI-CC82B9A8 recorded one stage earlier at trigger time (`queue/functions/self-upgrade.ts`) rather than inventing a new one.

**Soft blockers are not discounted at swap time.** BI-F36E7510 deliberately made the unattended scheduled path conservative — never drain while any work is in flight — and it stays that way. Only *observation* is excluded: every mutating (`ledger`) and external-reaching (`journal`) call still arms the blocker, and unclassified pre-Phase-3 rows (`auditClass IS NULL`) keep counting rather than being swept in by a bare `NOT` that SQL evaluates to NULL for them.

The policy lives in its own module (`activity-signal.ts`) rather than growing `quiescence.ts` past its size baseline, and the drain spec's B-class detection section (§6.4) now states the exclusion set as the contract.

## Verification — functional, not structural

- **(1)** A real child process is registered, `SIGKILL`ed, and a real `gate-worktree.mjs` launch against a mock MCP clears its record. The leaked record deliberately belongs to a different branch, sha, and session than the gate being launched.
- **(2)** A recycled pid — alive to `kill(0)` — is reaped via the **real** host process table, while a genuinely running gate survives past the TTL. A pre-fix record with no start time is bounded by the TTL; one inside the TTL still holds its slot.
- **(3)** The drain-blocker snapshot is empty while a gate polls, and still armed by a `ledger` row, a `journal` row, and an unclassified row. Driven through the real where-clause over an in-memory row store, so the assertion is about which rows survive the filter, not about its shape.

**Live evidence.** Over 24h on this install: of 105 five-minute windows that armed the blocker, **14 were armed by observation alone** and now clear; the other 91 still block.

**And it proved itself in production while gating this very PR.** The gate run recorded a `dead_queue_observers_swept` lease event clearing pid 50988 — a dead record belonging to *another session's* killed gate on `claude/local-ci-gate-readability-c32dac`. That session's live record, registered 30 seconds later, was untouched. Cross-session reach and the do-not-yank-a-live-slot property, both on the real shared queue.

## Notes for review

- Overlaps #4005 on `scripts/gate-worktree.mjs` (different concern — gate verdict readability; different region of the file). Whichever lands second rebases.
- The observer record gains one optional field; pre-fix records stay readable and are handled explicitly by the unverifiable path above.
- Not addressed here, tracked separately: the give-up path exiting 0, which is #4005's concern.

### DIFF
--- a/apps/web/lib/self-upgrade/activity-signal.test.ts
+++ b/apps/web/lib/self-upgrade/activity-signal.test.ts
@@ -1,0 +1,173 @@
+/**
+ * B-class activity signal — what counts as activity for the quiescence drain.
+ *
+ * Split from quiescence.test.ts: the exclusion set is a policy with its own
+ * module (activity-signal.ts) and its own failure history, and it deserves a
+ * suite that is read on its own.
+ */
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const prismaMock = vi.hoisted(() => ({
+  taskRunFindMany: vi.fn(),
+  taskRunUpdateMany: vi.fn(),
+  buildPhaseRunFindMany: vi.fn(),
+  buildPhaseRunUpdateMany: vi.fn(),
+  executeRaw: vi.fn(),
+  toolExecutionFindMany: vi.fn(),
+}));
+
+vi.mock("@dpf/db", () => ({
+  prisma: {
+    taskRun: {
+      findMany: (...args: unknown[]) => prismaMock.taskRunFindMany(...args),
+      updateMany: (...args: unknown[]) => prismaMock.taskRunUpdateMany(...args),
+    },
+    buildPhaseRun: {
+      findMany: (...args: unknown[]) => prismaMock.buildPhaseRunFindMany(...args),
+      updateMany: (...args: unknown[]) => prismaMock.buildPhaseRunUpdateMany(...args),
+    },
+    $executeRaw: (...args: unknown[]) => prismaMock.executeRaw(...args),
+    toolExecution: {
+      findMany: (...args: unknown[]) => prismaMock.toolExecutionFindMany(...args),
+    },
+    quiescenceRun: { findUnique: vi.fn(), update: vi.fn(), findMany: vi.fn() },
+  },
+}));
+
+vi.mock("@/lib/queue/inngest-client", () => ({ inngest: { send: vi.fn() } }));
+vi.mock("@/lib/tak/agent-event-bus", () => ({
+  agentEventBus: { broadcastSystem: vi.fn() },
+}));
+
+import { captureActiveSessionBlockers } from "./quiescence";
+
+beforeEach(() => {
+  vi.clearAllMocks();
+  prismaMock.taskRunFindMany.mockResolvedValue([]);
+  prismaMock.taskRunUpdateMany.mockResolvedValue({ count: 0 });
+  prismaMock.buildPhaseRunFindMany.mockResolvedValue([]);
+  prismaMock.buildPhaseRunUpdateMany.mockResolvedValue({ count: 0 });
+  prismaMock.executeRaw.mockResolvedValue(0);
+  prismaMock.toolExecutionFindMany.mockResolvedValue([]);
+});
+
+describe("B-class activity signal: a waiter must not count as activity (BI-2C7F51BA)", () => {
+  // The deadlock this fixes: a local-CI gate waits for the drain to clear, and
+  // the waiting is what keeps it armed. Every MCP call writes a ToolExecution
+  // row, INCLUDING read-only ones, so the gate's own get_quiescence_status poll
+  // — and the diagnostic call used to inspect the stall — re-armed the single
+  // soft blocker both were waiting on (observed: coordinator at ready-to-swap,
+  // one soft blocker, count 1).
+  //
+  // These drive the REAL where-clause through an in-memory row store, so the
+  // assertion is about which rows survive the filter, not about its shape.
+  type Row = {
+    toolName: string;
+    agentId: string;
+    auditClass: string | null;
+    createdAt: Date;
+  };
+
+  const NOW = new Date("2026-08-04T22:00:00.000Z");
+  const recently = (secondsAgo: number) => new Date(NOW.getTime() - secondsAgo * 1000);
+
+  /** A gate polling for admission, plus the diagnostic that inspects the stall.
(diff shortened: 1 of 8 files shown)


markdbodman:
Closing as a duplicate. Session `4251f277-e68d-48b8-8a54-a4fb884b06e4` was working the same backlog item concurrently and has its own fix on `claude/fix-bi-bugs-2c7f-f75a-da5056` (84f32f545d3); that branch is the one to ship. Operator call.

I reviewed theirs against mine before closing — it is complete and sound, and in a few places better: a `{alive, reason}` liveness verdict so a reap reports *why*, a TTL argued from the `--lease-wait-seconds` 7200 default, and a `process.uptime()`-derived start time that reads identically on every platform. No substantive coverage is lost by closing this PR.

Two things from this branch that are worth carrying over, since they are not in theirs:

**1. Live measurement of Defect 3's fix.** Over 24h on this install, of 289 five-minute windows, 105 armed `request.recent-tool-execution` before the change and 91 after — so **14 windows (13%) were armed by observation alone** and now clear, while the other 91 still block. That is direct evidence the narrowing is real *and* that it did not gut the signal. Query shape:

```sql
SELECT count(*) FILTER (WHERE old_cnt > 0) AS armed_before,
       count(*) FILTER (WHERE new_cnt > 0) AS armed_after
FROM (SELECT
  (SELECT count(*) FROM "ToolExecution" e WHERE e."createdAt" >= w.t - interval '5 minutes'
     AND e."createdAt" < w.t AND e."agentId" NOT LIKE 'edge-node:%') AS old_cnt,
  (SELECT count(*) FROM "ToolExecution" e WHERE e."createdAt" >= w.t - interval '5 minutes'
     AND e."createdAt" < w.t AND e."agentId" NOT LIKE 'edge-node:%'
     AND (e."auditClass" IS NULL OR e."auditClass" <> 'metrics_only')) AS new_cnt
  FROM generate_series(now() - interval '24 hours', now(), interval '5 minutes') AS w(t)) s;
```

Also confirmed live, for whichever exclusion mechanism ships: `get_quiescence_status` and `list_nonprod_environment_leases` are `metrics_only` (8,850 and 30,672 rows), while `claim_nonprod_environment_lease` and `record_local_integration_result` are `ledger` (74,143 and 855). The waiter's calls and the worker's calls do separate cleanly.

**2. Defect 1's startup sweep proved itself in production.** The gate run that greened this branch recorded a `dead_queue_observers_swept` lease event clearing pid 50988 — a dead record belonging to *another session's* killed gate on `claude/local-ci-gate-readability-c32dac`. That session's live record, registered 30 seconds later, was untouched. Both halves demonstrated on the real shared queue: the unfiltered sweep does reach across sessions, and it does not yank a live session's slot. Worth reproducing on the surviving branch before merge, since it is the property most likely to be "tightened" later.

One design note, take it or leave it: in `createObserverLivenessProbe` the TTL check runs *before* the start-time comparison, so a gate that is verifiably still running would be reaped at 12h. Ordering the positive proof first makes "verified live" outrank the TTL, leaving the TTL as the bound only when liveness is genuinely unprovable. At a 12h threshold this is theoretical — the argument from `--lease-wait-seconds` is sound — so it is a preference, not a defect.

Superseded work; branch deleted.

markdbodman:
Duplicate of concurrent work on claude/fix-bi-bugs-2c7f-f75a-da5056 — see the review handover above.

PR #4084, opened 2026-08-07
feat(federation): connection-readiness preflight on the pairing page (BI-CC8021CE)

markdbodman:
## What

Adds a **Connection readiness** checklist at the top of **Platform → Connections** (`/platform/federation-links`). It reads the install's environment server-side and shows three rows — exchange enabled, this installation's address (`PUBLIC_URL`), and local-network-peer allowance (`DPF_FEDERATION_ALLOW_INSECURE_PEERS`) — each as Ready / Action needed / Not needed. Every "Action needed" row prints **the exact `.env` line to add**.

## Why

Federation over HTTP-LAN needs three install-local `.env` values, and getting one wrong fails the pairing handshake with an opaque 401 / "base URL not configured" rather than a legible message. Observed live: an operator set `DPF_FEDERATION_EXCHANGE_ENABLED=1` but not `PUBLIC_URL` or `DPF_FEDERATION_ALLOW_INSECURE_PEERS`, and nothing on the page indicated the gap — so pairing would have silently failed. This turns a silent misconfiguration into a copy-the-line checklist, the first operator-legibility increment toward the zero-shell autodiscovery design.

The `lan-peers` row is context-aware: an HTTPS public address marks it *Not needed* (allow-insecure only matters for http / private-LAN hosts).

## Boundary

The card is **read-only**. It reports what's missing; it never edits the live install. Config is applied by the operator via `.env` + `/ops/self-upgrade` (AGENTS.md §1/line 68).

## Tests

- `apps/web/lib/federation/connection-readiness.test.ts` — 13 cases: all-set, each var missing, the exact real-world Mac state, https-public → not-applicable, RFC1918 boundary hosts (172.15/172.32 excluded), `NEXT_PUBLIC_APP_URL` precedence.
- `ConnectionReadinessCard.test.tsx` — renders "Ready to connect" vs "Setup needed" and surfaces the exact missing `.env` lines.
- Local-CI gate: passed (evidence `cmsigte2z023v01lalgtdn6tz`).

## Design grounding

Source of truth: `docs/superpowers/specs/2026-08-06-federation-zero-shell-autodiscovery-increment.md` + EP-MSP-FEDERATION B1 operator surface. Extends that surface with a read-only readiness view; no contract, schema, or route change. Operator doc updated (`docs/operations/federated-demand-channels.md`).


markdbodman:
Superseded: the readiness card still requires the operator to hand-edit `.env` and run a shell/ops upgrade — the exact manual step this work needs to eliminate. Pivoting to zero-typing auto-config: (1) resolve this install's federation authority URL from the request Host header instead of requiring PUBLIC_URL, (2) scoped same-org/private-LAN outbound allowance, (3) pairing enables exchange. New PR(s) to follow.

PR #4251, opened 2026-08-13
fix(inference): keep Anthropic prompt cache alive across tool failures (BI-56804810)

markdbodman:
## Summary
On the Anthropic cloud route, prompt caching stopped paying off for the rest of most sessions. Anthropic caches its prefix in the canonical order **tools → system → messages**, so the stable system-prefix cache breakpoint (`apps/web/lib/routing/anthropic-cache.ts`, BI-79A5C00F) also covers the **tools** block. The agentic loop called `enrichToolDescriptions` **every turn**, appending `[WARNING …]` / `[REVIEW REJECTION …]` strings **into tool `description` fields** once any tool failed or was review-vetoed. That mutated the tools block on nearly every subsequent turn → the cached tools+system prefix was busted and re-billed at full input rate (vs ~0.1× cache reads) for the remainder of the session.

**Fix (zero capability cost):** relocate the per-turn session tool-notes out of tool descriptions and into a synthetic user message appended to the **messages tail** — *after* the cached prefix in Anthropic's ordering — via the new `buildToolSessionHintMessage`. The model still gets the identical "don't blindly retry" signal, but the tools block stays byte-identical turn-over-turn so the existing breakpoint actually gets cache hits. Failure/veto semantics (sticky review veto, clear-on-passing-review, per-tool error warning, no-op when clean) are unchanged and re-tested against the message form.

Backlog item: **BI-56804810** (portfolio/bug, EP-CLAUDE-INSIDE-OUT, size small, triaged build).

## Scope
Anthropic cloud route **only**. The local served-model path (DMR/llama.cpp, OpenAI-compatible) never sees `cache_control` and is unaffected. Delivery via the messages tail is provider-agnostic, so no per-provider branching.

## Verification
- `apps/web/lib/tak/agentic-loop.test.ts` + `apps/web/lib/routing/anthropic-cache.test.ts`: **121 passed**.
- **Functional cache-key proof** (new): the `tools`+`system` cache-prefix surface is **byte-identical** across a tool failure under the fix (cache hit); a regression guard documents that the retired in-description mutation **changed** that surface (cache miss) — so the exact bytes Anthropic hashes to gate `usage.cache_read_input_tokens` are now stable.
- Typecheck clean (`run-tsc --noEmit`, 8 GB heap); tool-description-hygiene guard passes; module-size ratchet re-baselined; local-CI merged gate **passed** for this HEAD.
- **Remaining step (needs an Anthropic-configured provider):** the live `usage.cache_read_input_tokens` before/after A/B — drive a multi-turn loop that induces a tool failure and assert cache reads stay ~full on later turns instead of dropping to ~0. Cannot run in this session (no live Anthropic inference); tracked on the BI.

## Secondary (measure, not changed here)
Progressive tool loading (`load_tools`, auto-load-on-direct-call) grows `routeOptions.tools` mid-loop, which also mutates position 0 and invalidates the prefix on the next turn. On Anthropic, front-loading the full authorized tool set once may net cheaper for long loops — worth a measurement; out of scope for this fix. Noted on the BI.

## Design grounding
- Specs/plans reviewed: `docs/superpowers/specs/2026-06-20-context-engineering-tool-efficiency-design.md` (line 81 confirms the cached prefix is *tools+system*; line 181 documents the `enrichToolDescriptions`-onto-descriptions pattern as a P11 retention win, never reconciled against P8 prefix stability — this PR resolves that tension), `docs/superpowers/specs/2026-05-19-ai-cost-governance.md`.
- Source of truth: `docs/architecture/context-engineering-standards.md` P8 ("Cache the stable prefix") + the `anthropic-cache.ts` breakpoint contract (BI-79A5C00F).
- Decision: move the sticky failure/veto signal from tool descriptions (kept P11, broke P8 for tools) to the messages tail — preserving P11 while restoring P8. Delivery channel only; no spec/schema/tool-surface contract change.
- `Design-Grounding-Decision: no-contract-change`

## Notes
- DCO: commit is `Signed-off-by` (dpf-ci).
- The pre-push semantic-review-gate reported `missing-receipt` as a **shadow observation** (non-blocking; publication allowed). An independent semantic-review receipt for the stable commit is still owed per DPF process before merge.

### DIFF
--- a/apps/web/lib/queue/functions/self-upgrade.test.ts
+++ b/apps/web/lib/queue/functions/self-upgrade.test.ts
@@ -41,6 +41,11 @@ const mocks = vi.hoisted(() => ({
   getCooldownUntil: vi.fn().mockResolvedValue(null),
   recordCooldown: vi.fn().mockResolvedValue(undefined),
   clearCooldown: vi.fn().mockResolvedValue(undefined),
+  // Host-memory-headroom guard (BI-EFA383AA): mock so the unit test never reads
+  // real host memory. The real guard calls os.freemem(), which under-reports on
+  // macOS (cache-heavy) and defers the whole suite on any memory-constrained
+  // host. Default to "enough headroom"; the defer path is exercised via override.
+  evaluateHostMemoryGuard: vi.fn().mockResolvedValue({ defer: false }),
   emitUpgradeEvent: vi.fn(),
   createSelfUpgradeRecoveryPoint: vi.fn(),
   summarizeRecoveryPointFailure: vi.fn(),
@@ -155,6 +160,14 @@ vi.mock("@/lib/self-upgrade/cooldown", () => ({
   isInCooldown: (until: Date | null, now: Date) =>
     !!until && now.getTime() < until.getTime(),
   DEFAULT_COOLDOWN_MINUTES: 30,
+}));
+
+// BI-EFA383AA: override only evaluateHostMemoryGuard (real impl reads os.freemem);
+// keep every other export real via importActual so formatGiB / checkHostMemoryHeadroom
+// / constants are unaffected.
+vi.mock("@/lib/self-upgrade/host-memory-preflight", async (importOriginal) => ({
+  ...(await importOriginal<typeof import("@/lib/self-upgrade/host-memory-preflight")>()),
+  evaluateHostMemoryGuard: mocks.evaluateHostMemoryGuard,
 }));
 
 vi.mock("@/lib/self-upgrade/notifications", () => ({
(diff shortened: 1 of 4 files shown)


markdbodman:
Rebased onto current `main` (a74c1a45) — resolves the merge conflict with sibling #4250 (both touched `agentic-loop.ts`); confirmed both changes coexist and re-verified (122 tests, typecheck clean).

Also bundled a second commit fixing **BI-EFA383AA**: `self-upgrade.test.ts` called the real `evaluateHostMemoryGuard`, which reads `os.freemem()` and defers on memory-constrained/macOS hosts (~2.3GB free on a 128GB Mac), making the whole-tree local-CI pre-push gate unrunnable on macOS (49 false failures). Linux CI reads `/proc/meminfo` and was unaffected, which is why GitHub stayed green. Mocked the guard (importActual-preserving other exports) with host-independent defer/proceed coverage. Included here because the local gate is otherwise unrunnable on this host, blocking the cache fix from passing its own gate. Local-CI gate now passes @ 981db1d74.

markdbodman:
Cleared the last red guard: **Docs Impact Gate** (a PR-only guard not in the local preflight) flagged `agentic-loop.ts` as linked to `docs/design/golden-triangle-design.md`. This change has no user-visible behavior change — session tool-notes move from tool descriptions to the messages tail, and the model still receives the identical signal — so I added a `Docs-Impact-Decision:` attestation trailer (no user-guide update needed). Re-gated locally (full suite green @ 164649cff) and pushed.

PR #4314, opened 2026-08-15
fix(deps): raise nanoid floor to 3.3.18 (GHSA-2v37-7h3g-55p8)

markdbodman:
Fixes a **stale security floor**: the existing `nanoid@<3.3.17` override no longer clears **GHSA-2v37-7h3g-55p8** / CVE-2026-67213 (high).

## Why this was invisible

The advisory was **revised after our floor landed**. It originally described predictable output for a non-integer size, and we floored to `^3.3.17` (Dependabot #152). It has since widened — custom generators can loop indefinitely when `size` is zero — and the patched releases moved to **3.3.18 / 5.1.6**.

Because Dependabot #152 was already closed by the original floor, **no new alert fired**, and the tree quietly kept resolving the now-vulnerable **3.3.17**. This was surfaced by the OSV dependency scan (`pnpm scan:deps`) while working on the `extract-zip` fix, not by the Dependabot feed.

Worth noting as a class of gap: *a closed alert does not stay fixed if the advisory's patched range moves.* The `Stale override audit` check is the natural home for catching this.

## Change

- Raise the 3.x floor: `'nanoid@<3.3.17': '^3.3.17'` → `'nanoid@<3.3.18': '^3.3.18'`, with the override comment updated to record the revision.
- The 5.x line already resolves **5.1.16** (> the 5.1.6 patch), so it stays unfloored — preserving the existing scoping intent.

## Verification

- Lockfile regenerated via `scripts/regen-lockfile.mjs`; **exactly one package changed** (`nanoid` 3.3.17 → 3.3.18), reported **SCOPED** and **STABLE**.
- Lockfile now resolves `nanoid` 3.3.18 and 5.1.16 — no vulnerable copy remains.
- `pnpm scan:deps` no longer reports `nanoid`.
- Override Provenance and lockfile release-age guards pass (3.3.18 published 2026-08-07, well clear of the 1440-minute floor).
- **Local CI gate passed** on the exact tree (`eac75c063aee`).

## Relationship to the other PR

Independent of #4313 (`extract-zip`). Both touch `pnpm-workspace.yaml`'s overrides block in different places, so whichever merges second will need a trivial lockfile regen if git conflicts.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,7 +17,7 @@ overrides:
   node-forge: 1.4.0
   postcss: ^8.5.18
   uuid: ^14.0.0
-  nanoid@<3.3.17: ^3.3.17
+  nanoid@<3.3.18: ^3.3.18
   dompurify: ^3.4.13
   '@opentelemetry/otlp-transformer>protobufjs': ^8.3.0
   protobufjs@<7.6.5: ^7.6.5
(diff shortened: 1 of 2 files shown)


markdbodman:
Folded into #4313. The `OSV vulnerability scan` gate is whole-tree at `--fail-on high`, so this PR failed on #4313's `extract-zip` finding and #4313 failed on this one's `nanoid` finding — neither could go green alone. Both fixes now ride together in #4313, with the nanoid commit kept intact.

PR #4340, opened 2026-08-15
refactor(workroom): converge the WorkRoom view vocabulary onto Workroom (BI-C2C16582)

markdbodman:
Phase 2 of the Workroom canonical rename (`EP-WORK-CONVERGENCE`, umbrella `BI-D2D190BF`). Phase 1 renamed the record; this merges the **other** vocabulary that already owned the name, so there is one stem instead of two spellings of the same word.

## What changed

One stem, one casing — `Workroom`, never `WorkRoom` / `workRoom` / `WORK_ROOM` / `work-room`.

61 code files: `WorkroomView`, `WorkroomCycle`, `WorkroomOutcomePacket`, `WorkroomParticipantView`, `WorkroomStructure`, `authorizeWorkroomAccess`. Component directory and files moved with `git mv` so history follows. Doc prose converged in 5 live pages; dated specs, plans and ux-fit records keep their original wording as the record of what was decided when.

## The actual deliverable: the boundary

`docs/architecture/workroom-vocabulary-boundary.md`. **A Workroom is claimed as a record, projected as a WorkCase, and rendered as a Workroom view.** View types keep a role suffix (`WorkroomView`, `WorkroomCycleView`) because that suffix is now the only thing distinguishing a projection from the record.

## A tension recorded, not hidden

`WORK_CASE_SOURCE_REGISTRY` carries **13 sources**, each with its own `roomProjection`: `task-node`, `backlog-item`, **`work-capsule`**, `approval`, `data-control-operation`, `manual-task`, `scheduled`, `engagement`, `opportunity`, `booking`, `storefront-booking`, `activity`, `field-service-job`.

So the room is a *container* and the claim record is *one of the 13*. Converging means one source shares the container's name, and a booking or field-service job renders in something called a Workroom. **This was raised before landing and the direction was reaffirmed by the operator.** The boundary doc names the disambiguation rule ("the Workroom record" / "the WorkCase" / "the Workroom view") and the reversible exit — rename the container, not the record — if it proves costly.

## ⚠️ Guard exemption — NEEDS REVIEW

The UX-Fit gate listed added `.tsx` with `--diff-filter=A` and **no rename detection**, so `git mv` read as delete+add and every moved component was reported as "adds a user-facing control", demanding measured UX evidence for screens nobody changed.

I did **not** supply a manifest for a non-change, and did not invent a `decisionInteractionId` or sweep numbers — that is the attestation theatre `BI-D967DEE0` rebuilt this gate to stop.

My first attempt was a 95% similarity threshold. **That was wrong**: git scored the moved files 90–97% because renaming a repeated identifier drags similarity down even when nothing changed, so the threshold would have been tuned to my own diff. Verified exactly instead — *for every moved component, the old blob with the rename substitution applied equals the new file byte-for-byte.*

The shipped rule compares what the gate cares about: user-visible copy (quoted strings and JSX text), ignoring module specifiers, compared normalized (a rename re-cases and re-spaces the same words). **Text that survives normalization as new IS new copy and still requires measured evidence.**

Three fixtures pin it — a re-casing rename is exempt, **a rename that ADDS a control is NOT exempt**, and a plain added file is never treated as a rename. That middle test is load-bearing: it stops the exemption becoming a way to smuggle UI in behind a file move.

Prose-lint baseline keys were renamed **in place with recorded values untouched** — still 1205 files, no ratchet loosened.

## Verification

`tsc --noEmit` clean · 404 tests across 69 files · 26 gate tests · derived artifacts regenerate cleanly (448 code edges, 87 route edges) · local-CI gate passed, evidence `cmsv0kfcs000i01mrpuawi971`, gated sha `4875a02dff57`.

The `doc-impact` generator caught the moved component paths in the user guide's `relatedCode` frontmatter; repointed rather than regenerated around, so doc→code edges stay real.

## Still deliberately unrenamed

MCP tool names behind their alias window (`BI-0702869B`) and the Prisma field vocabulary — `workCapsuleId`, the `capsule`/`workCapsule` relation fields, and the `capsuleId` semantic key with its `WC-*` values, whose columns are unchanged on disk under `@@map`.

### DIFF
--- a/apps/web/app/(shell)/workspace/cases/[caseKey]/page.tsx
+++ b/apps/web/app/(shell)/workspace/cases/[caseKey]/page.tsx
@@ -5,8 +5,8 @@ import { WorkCaseDetailView } from "@/components/workspace/WorkCaseDetailView";
 import { auth } from "@/lib/auth";
 import { getGrantedCapabilities } from "@/lib/permissions";
 import { loadEffectiveAuthContext } from "@/lib/identity/load-effective-auth-context";
-import { loadPrismaWorkRoomParticipants } from "@/lib/work-management/room-participation-prisma.server";
-import { resolveWorkRoomStructureForCase } from "@/lib/work-management/room-structure.server";
+import { loadPrismaWorkroomParticipants } from "@/lib/work-management/room-participation-prisma.server";
+import { resolveWorkroomStructureForCase } from "@/lib/work-management/room-structure.server";
 import { loadWorkspaceWorkCaseDetail } from "@/lib/work-management/workspace-case-loader";
 
 type Props = {
@@ -36,8 +36,8 @@ export default async function WorkspaceCaseDetailPage({ params }: Props) {
       sensitivityClearance: effectiveAuth.sensitivityClearance,
       isSuperuser: effectiveAuth.isSuperuser,
     },
-    participantLoader: loadPrismaWorkRoomParticipants,
-    structureLoader: resolveWorkRoomStructureForCase,
+    participantLoader: loadPrismaWorkroomParticipants,
+    structureLoader: resolveWorkroomStructureForCase,
   });
   if (!detail) notFound();
 
--- a/apps/web/components/ui/report-kit/statusColors.test.ts
+++ b/apps/web/components/ui/report-kit/statusColors.test.ts
@@ -99,9 +99,9 @@ describe("statusColors", () => {
 
   it("maps Work Room state, outcome health, and activity through shared domains", () => {
     expect(resolveIntent("workCaseState", "waiting-on-person")).toBe("warning");
-    expect(resolveIntent("workRoomOutcomeHealth", "at-risk")).toBe("warning");
-    expect(resolveIntent("workRoomActivity", "message")).toBe("neutral");
-    expect(resolveIntent("workRoomActivity", "decision-resolved")).toBe("success");
+    expect(resolveIntent("workroomOutcomeHealth", "at-risk")).toBe("warning");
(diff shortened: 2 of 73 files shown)


markdbodman:
**The UX-Fit exemption in this PR is a stopgap, and is now tracked as such.**

A larger UX rework is underway (operator, 2026-08-15), so I've filed `BI-EDDD4E44` rather than let this quietly become permanent. It records:

- **Why it exists** — `--diff-filter=A` with no rename detection makes `git mv` read as a new user-facing control.
- **Why it's a stopgap, not a design** — the copy extractor is regex over diff lines, so it *approximates* user-visible text instead of deriving it from the rendered DOM, which is what the UX budget actually measures. And it reasons about a diff, when this gate's own thesis (`BI-D967DEE0`) is that measured artifacts beat inferred ones.
- **A known over-reach** — normalized comparison treats `"Work Room"` → `"Workroom"` as not-new. Correct for a rename pass; wrong as a general rule, since it would also excuse a genuine copy edit that only re-cases or re-spaces.
- **Suggested direction** — decide rename-vs-new-surface from the committed route-budget baseline (keyed by *route*, not file path), so moving a component cannot change the verdict and the diff-level heuristic disappears entirely.

Same path-keying weakness exists in the prose-lint baseline: this PR renamed those keys in place with values untouched (1205 files, no ratchet loosened), but it will recur on any future move. Also folded into that BI.

Reviewers: the negative fixture is the one to check — *a rename that ADDS a control is NOT exempt*. That's what stops this becoming a way to smuggle UI in behind a file move.

PR #4406, opened 2026-08-18
fix(pregate): wire the pregate script family to the symlink-robust entry guard

markdbodman:
## Problem

The Janitor Tests bundle fails on macOS hosts in any worktree, including the untouched root clone: the refusal contract test ("gate-worktree.mjs refuses to run when neither an explicit command, the stub, nor the checked-in Node runner exists", tests/release/pregate-node-gate-contract.test.mjs) expects a non-zero refusal but gets exit 0 with zero output.

Root cause (BI-745658D7): the pregate script family — `scripts/gate-worktree.mjs`, `scripts/pregate.mjs`, `scripts/pregate-status.mjs`, `scripts/local-ci-runner.mjs` — gates `main()` on `process.argv[1] === THIS_FILE`. Node realpath-resolves the ESM entry module, so under a symlinked invocation spelling (macOS tmpdir `/var/folders/…` → `/private/var/…`, symlinked checkouts) `import.meta.url` names the realpath while `argv[1]` keeps the caller's spelling. The guard is false, `main()` never runs, and the process exits 0 silently — a **false pass** from a gate whose whole job is to refuse. Confirmed by A/B probe: the identical temp-tree copy invoked via its `/private/var` realpath refuses correctly with exit 1.

## Fix

Completes the `isEntryModule` adoption that #4404 started. That sweep landed `scripts/lib/entry-module.mjs` (helper + unit test + Janitor Tests registration) and converted every other entry script, but deliberately left these four to this branch:

- Wire all four pregate-family scripts to `isEntryModule(import.meta.url)` (realpath comparison, exact-match fast path preserved); drop now-unused `THIS_FILE`/`fileURLToPath` bindings where nothing else needs them.
- Add the missing regression test: invoke the gate through a directory symlink (junction-flavored so it also runs on Windows without symlink privilege) and assert `main()` actually produces the dry-run output. Linux CI cannot catch this trap — its tmpdir is not a symlink — so this test is the only cross-platform lock.
- Add platform-support-watchlist row **S10** so the next entry script does not reintroduce the naive guard; regenerate `doc-impact.generated.json` accordingly.

## Verification (macOS host — the only place the defect reproduces)

- `node --test tests/release/pregate-node-gate-contract.test.mjs scripts/lib/entry-module.test.mjs tests/release/local-ci-gate-contract.test.mjs` → **62/62 pass**, including the previously-failing refusal test and the new symlinked-invocation test.
- `node scripts/ci-policy-guards.mjs --profile source` → all **48 guards pass** (includes the full Janitor Tests bundle).
- A/B: pre-fix gate invoked via a symlinked path exits 0 silently; post-fix gate runs `main()` and prints the dry-run output — the new test's stdout assertion pins exactly this.
- Semantic review receipt recorded for `WC-A1FD21C9` at the exact committed tree (`pass`, evidence `cmszats0r00j301pj3xafr4vg`).

Backlog: BI-745658D7 (triaged build, small) · EP-413F2602 · supersedes the pushed branch `fix/pregate-entry-guard-symlink` (pre-#4404 version of the same fix; will be deleted after merge).

Local-CI-Evidence: cmszawh4u00jt01pjar2immgn (fix/pregate-family-entry-guard@f579afec64d5a72c97136dca061e048c41969de1)
Docs-Impact-Decision: internal entry-guard robustness fix in gate scripts; every documented pregate command, flag, and workflow in docs/testing/pre-pr-gate.md is unchanged — the contributor-facing delta (watchlist row S10) ships in this PR


markdbodman:
macOS verification pass (independent second signal, from the session that filed BI-745658D7): checked out the exact PR head f579afec6 into a fresh scratch worktree on the macOS host where the defect reproduces (Darwin 25.5.0, /var→/private/var symlinked tmpdir — the environment Linux CI cannot exercise).

- `node --test tests/release/pregate-node-gate-contract.test.mjs scripts/lib/entry-module.test.mjs` → **31/31 pass, 0 skipped** — including the previously-failing refusal contract test ("refuses to run when neither an explicit command, the stub, nor the checked-in Node runner exists") and the new symlinked-invocation regression test, which executed (not skipped) on this host.
- `node --test tests/release/local-ci-gate-contract.test.mjs` → **31/31 pass**.

Before this branch, the refusal test failed on this host at main 1b2bf3d99 (gate exited 0 with zero output because the naive argv[1] guard never ran main()). With this branch it refuses correctly. Matches the author session's own 62/62 evidence.

PR #4469, opened 2026-08-22
feat(mileage): mileage absorption data substrate (BI-6D98AD8A, BI-E17E0034)

markdbodman:
## What

The native data substrate for MileIQ-class mileage tracking, absorbed onto the existing `ExpenseClaim` reimbursement spine rather than a parallel path. Six models, one purely-additive migration.

| Model | Job |
|---|---|
| `Vehicle` | Company vehicles bind to their `FixedAsset`, so depreciation and mileage share one asset identity |
| `Trip` | One captured drive; distance in integer metres so no rounding is baked in at capture |
| `TripClassificationRule` | Repeated-route, work-hours and commute-exclusion in one model — they differ by predicate, not by shape |
| `DriverLocationConsent` | Capture precondition; pins the policy version so a changed disclosure needs fresh consent |
| `MileageRatePlan` / `MileageRate` | Effective-dated — a trip prices at the rate in force on the day it was driven, never today's rate |

## Why this shape

Privacy is structural rather than procedural. `Trip` geometry is precise personal location for an identified employee, so it is consent-gated, masked on read with `collectionRule: "minimize"`, and exempt from row purge because a trip that priced a reimbursement is accounting evidence. `TripClassificationRule` gets the same treatment — a commute-exclusion predicate can embed a home location, so it is location data wearing configuration's clothes. Revoked consent is retained, not deleted: the consent history is the lawful-basis evidence for every captured trip.

Effective-dating the rate is the other load-bearing choice. Without it a rate change silently re-prices history and an IRS-compliant mileage log stops being reproducible.

## Verification

- Migration is purely additive (no `DROP`, no column alteration) and the **full chain applies cleanly from empty** against a scratch Postgres — all six tables created
- `pregate` sandbox gate **PASS** at `79a2dd4dc199`, evidence `cmt4hqe4g1jm001qwlgpsh8kn`
- `pregate:preflight` — 47 guards clean
- Closed-set-strings, FK-index-coverage and not-active-convention ratchets pass
- 7 shape tests in `packages/db`; coverage + merge-completeness suites green

## Gate cascade closed properly

Two gates the sandbox run caught were real defects, not noise:

1. A `Trip` attributed to a customer had no merge repoint step — merging two accounts would have stranded job-costed mileage on the tombstoned loser.
2. The six models were unregistered in the data-governance asset registry.

Classification, retention disposition, and the data-impact manifest are all declared rather than baselined. No model was added to a grandfather baseline.

## Honest notes

- `DriverLocationConsent.retentionDays` **declares** the geometry retention bound; the field-level minimisation job that enforces it is a named follow-up and is **not** shipped here. Minimisation must clear geometry columns, never delete rows.
- The governance taxonomy has no `location` `DataCategory`, so precise personal location registers as `personal-attribute`. Worth revisiting as a taxonomy gap rather than leaving location semantics implicit.
- The plan doc lands here carrying a correction: its first pass was written against a stale branch and understated payroll. `PayRun`, `Payslip` and a manual NACHA disbursement rail already ship on `main`; BI-EAC670F1 and BI-AC74BE73 have been rescoped in the backlog accordingly.

Refs: EP-MILEAGE-ABSORB, BI-6D98AD8A, BI-E17E0034, kernel decision DI-87A27C692B16

Docs-Impact-Decision: schema-and-registration only — this PR adds six models, their classification and retention disposition, and a substrate baseline line. It ships no route, no UI and no user-visible behaviour, so no user-guide page changes. The flagged architecture runbooks (backlog-and-planning, enforced-ci-gates, platform-substrate-boundaries) describe the gate process itself, which is unchanged: this PR follows those gates rather than altering them. User-guide coverage for mileage lands with the surface that exposes it.
Local-CI-Evidence: cmt4l0zfk1vik01qwd5v27nvc (feat/absorption-data-substrate@77c19e769520)


markdbodman:
Superseded by #4481. This PR's record was left pinned to an unreachable head after a force-push that fixed a DCO authorship mismatch, and GitHub would not reopen it. Same branch, same commits, same passing gate — see #4481.

PR #4602, opened 2026-08-23
feat(tax): seed the US federal employment-tax authority and state payroll types (BI-947F8703)

markdbodman:
Payroll tax had no authority to file against. The jurisdiction reference seeded all 50 states plus the EU and a GB override, and **no national authority** — so a Form 941 had nowhere to go. This is items 1 and 2 of BI-947F8703.

| | |
|---|---|
| `TAX-JUR-US-FEDERAL` | Internal Revenue Service — `payroll_withholding`, `fica`, `futa`; cadence includes `semiweekly` |
| US states | gain `suta`, and `payroll_withholding` **only where the state actually withholds wage income tax** |

## Wage withholding is not universal

Nine states are excluded: **AK, FL, NH, NV, SD, TN, TX, WA, WY**. New Hampshire taxes only interest and dividends; Washington only capital gains — neither withholds from wages.

This matters more than a data nicety: the tax spine turns a declared obligation into a `TaxObligationPeriod` that comes **due**. Seeding `payroll_withholding` for Texas would put a recurring filing obligation on an install that does not owe one, and the operator would see a period they cannot file.

Every state keeps `suta`, because every state runs unemployment insurance. Asserted in tests: CA has withholding, TX and FL do not, all four have SUTA, and `sales_tax` is untouched.

## The federal filing note carries the expensive part

Deposit cadence is **assigned** from a lookback at prior liability, not chosen. Depositing on the wrong cadence attracts a penalty even when the amount is right — so the note says so rather than leaving an operator to discover it.

## URLs were verified, not asserted

`irs.gov/businesses/small-businesses-self-employed/employment-taxes` and `eftps.gov` were both fetched and confirmed live; the IRS page names 941, 940 and W-2 as expected.

The check also surfaced something worth recording: the IRS is migrating **individual** taxpayers off EFTPS (new individual enrollments stopped 2025-10-17), which does **not** affect employment-tax deposits. That is in the filing notes so nobody later reads the migration notice as a reason to change the business deposit path.

`confidence` stays `medium` and `lastVerifiedAt` is **unset** — the URLs resolve, but nobody has walked a real filing through them. For a spine that schedules money movement, "reachable" and "proven" are not the same claim.

## Verification

- `pregate` **PASS** at `21619a479dff`, evidence `cmt66yc1t0cld01mqb920dm4i`
- `pregate:preflight` — 50 guards clean
- 2049 db tests pass

## Still open on BI-947F8703

Generalizing `TaxObligationPeriod`'s sales-shaped amount fields, wiring the `payroll_run` emitter to persist snapshots, `TaxDepositSchedule`, and the 941/940/W-2 generators. The rate figures themselves remain an operator-supplied, source-cited task (BI-4EB27955).

Refs: BI-947F8703, EP-PAYROLL-ABSORB

Local-CI-Evidence: cmt69vjad0nzu01mq0n41w7c2 (feat/payroll-tax-jurisdictions@88ea4cde7a5d)


markdbodman:
**Blocked on BI-E411EBEC by kernel ruling `DI-F3554F482E80`** (composite 10.225, margin 0.460, high confidence, verdict proceed).

The jurisdiction filter in this PR is correct but does **not** clear the sweep. `jurisdictionCountryScope()` deliberately returns no filter when `OrganizationTaxProfile.homeCountryCode` is unset, and the seeded install never sets one — so the default install, which is what CI measures and what a fresh customer starts with, still renders all 81 authorities. The sweep still reports `choices in one control: 80 → 81`.

That confirms what the earlier tie (`DI-89F317F406AA`, 9.118 vs 8.926) was actually signalling: the filter and the picker are **complements, not alternatives**. The filter fixes a correctness defect — a US business should never have been offered EU VAT authorities. The picker fixes the control. Neither substitutes for the other.

Two alternatives were weighed and rejected:
- **Build the picker in this PR** — 9.765. Widens a seed-data PR into a shared UI primitive change.
- **Give the seeded install a home country** — 4.936, decisively rejected; negative on "Outbound and irreversible actions require explicit go" and flat on "Preserve the model; disclose complexity progressively". It would mask the unfiltered path rather than fix the control, which stays 4× over budget at ~51 options.

This PR holds until BI-E411EBEC lands. Its own content — the federal tax authority, state payroll types, and the scoping filter — is complete and gate-passed (`cmt69vjad0nzu01mq0n41w7c2`).

Also on this run: `Unit Tests (web shard 2/4)` failed on `RestaurantFloorOperations.test.tsx`. This PR touches five files, none under `twin/restaurant`, and that test passes 3/3 locally on this branch — it looks like shard pollution rather than a defect here.

PR #4755, opened 2026-08-27
feat(mcp): let external agents act as a coworker so they can join Work Rooms (BI-B986A18B)

markdbodman:
## The room was never closed by policy. It was closed by omission.

Your Work Room design has been built since 2026-08-12 — `post_room_message`, `read_room_messages`, `invite_room_participant` (agents **and** people), `resolveAgentRoomAccess`, the `coordinator` role, join-on-post presence. All shipped.

I tried to use it. It said:

```
invalid_caller — "post_room_message requires an acting coworker."
```

Every room handler resolves its caller from `context.agentId`, which the MCP route reads off the bearer token:

```ts
// apps/web/app/api/mcp/v1/route.ts
agentId: token.agentId ?? undefined,
```

**Every `McpApiToken` row in the database has `agentId = NULL`.** Every one ever issued, including all the `codex-*` tokens.

So no external CLI agent — Claude, Codex, Grok — has ever been able to join, post to, or read a Work Room. Three hardcoded nulls and a missing dropdown:

| `apps/web/lib/actions/mcp-tokens.ts` | |
|---|---|
| `issueMyMcpToken` | threads `agentId` through — the plumbing was already correct |
| `issueMyTemplateMcpToken` | `agentId: null` |
| `rotateMyMcpTokenWithEdit` | `agentId: null` — **rotation silently demoted a room-capable token to anonymous** |

`McpTokenManager` offered no way to pick a coworker, so the one path that *did* support binding was unreachable. And the registry had nothing to bind to: all 73 registry agents were in-platform coworkers.

## What it cost

**6 `WorkItemMessage` rows exist in total. Against 604 `TaskMessage` rows.**

One external thread ran ~58 hours over two days and sent **235 of its ~284 coordination messages** through its vendor's own task bus — peer-to-peer injections into other agents' context windows — because the room it should have used refused it. That is the concrete argument for §0 of the design, and it is why this is worth fixing.

## What changed

- **Registry identities** — `AGT-EXT-CLAUDE`, `AGT-EXT-CODEX`, `AGT-EXT-GROK` seeded with `work_room_read` / `work_room_write` / `registry_read` (fix the seed, not the runtime).
- **Binding on issue**, and **preserved across rotation** — rotation changes the secret, never the identity.
- **An "Acts as coworker" control**, defaulting to *"None — cannot join Work Rooms"*, listing only coworkers that hold `work_room_write`.
- **The binding rendered on every token row**, so an anonymous token is visible instead of silent. That invisibility is how this survived.
- `listActingCoworkerOptions` — active agents only, deduped, sorted.

## What did NOT change

**Binding grants identity, never admission.** A bound token is *someone*; it is not thereby *in* any room. `authorizeWorkRoomAccess`, outcome-scoped membership and the clearance ceiling are untouched — a coworker cleared for an HR room is still not cleared for a finance room. An unbound token is still refused. Nothing becomes fail-open.

## Evidence

- **59/59** tests in the token action suite, including five new guards: binding applied on issue, absent when unchosen, **preserved across rotation**, offer list filtered to active room-capable coworkers and deduped, nothing offered to an unauthenticated caller.
- Typecheck clean; guard-parity preflight **52/52** on this exact tree. The governed local-CI sandbox gate did **not** reach a verdict — see *Gate status* below; do not read this line as a sandbox pass.

## Review notes — two things stated plainly rather than buried

**1. The UX decision is uncertain, not ratified.** `principle_decide` returned `DI-D55ECEB2DA17` with **margin 0.000 across all three options**, `confidence: low`, `autonomyEligible: false`, no commandment conflict. The scorer did not discriminate. I took the most conservative option — the only one where a room-capable identity cannot be granted without the operator seeing and choosing it — and recorded it in `docs/ux-fit/2026-08-27-mcp-token-acting-coworker.ux-fit.json` as uncertain. **Overrule toward auto-binding freely**; the server action and the binding are unchanged by that choice, only the control is.

**2. One pre-existing repair, surfaced not chosen.** The MCP authorization runbook cited `BI-86EF5900` for the code-graph trust-vector rule. Neither that id nor `BI-F9CAF214` — the BI of the PR that introduced it (#4589) — resolves in the live backlog; both predate a backlog reset. The rule is real and already shipped, so filing a new BI for delivered work would be wrong. The citation now points at source behaviour instead. Touching that file is what exposed it.

Module-size baseline moved for two files (`McpTokenManager.tsx` 1326→1383, `mcp-tokens.test.ts` 1254→1369), justified in the re-baseline commit.

## Gate status — stated plainly

The governed local-CI sandbox gate **could not produce a verdict** on this tree across two runs, and neither failure is about this diff:

1. Fenced mid-run by `portal_quiescing` — a real self-upgrade was draining, and its only soft blocker was `request.recent-tool-execution`, i.e. the agent waiting on the gate.
2. `canonical local-CI executor ended without publishing evidence before the deadline`, then `nonprod_lease_not_owner` on release — the single `local-integration-ci` slot had been taken by another concurrent session.

`pregate:status` reports `gate record status failed with NO recorded reason ... Re-run pregate` in both cases, which is the tool correctly saying *contended slot, not a verdict*.

Three agent sessions were competing for one CI slot at the time (`BI-D908DA0A`). Rather than fire a third run, I stopped: two failures of one blocker class is a stop-and-hand-back under the descent bound merged in #4737.

**What IS verified on this exact tree:** guard-parity preflight 52/52, typecheck clean, 59/59 unit tests including the five new binding guards. **GitHub CI is the authoritative gate here** — auto-merge is deliberately NOT armed. Please let the checks decide.

## Design grounding

Extends `docs/superpowers/specs/2026-08-12-work-room-multi-agent-communication-substrate-design.md` (EP-WORKROOM-COMMS; kernel `DI-A92A7184020F` intra-first, `DI-1C65C5205848` coordinator-as-first-class), which gains a §9 delivered-state section recording what shipped, what this closes, and what remains. Operator-facing detail lives in `docs/architecture/mcp-tool-authorization-runbook.md`, which `AGENTS.md` §6 already points at — deliberately kept out of `AGENTS.md` itself so the always-on instruction plane does not grow.

## Still open (deliberately)

Room-keyed subscribe/push — §4's "agent read/subscribe" is served by pull today. Agent presence beyond join-on-post. The participant relation and multi-holder occupancy, which is `BI-D4C110BC` (reactivated from deferred with four measured gaps). All of Phase 2 / GAID federation (§5).

Backlog item: `BI-B986A18B`


markdbodman:
## UX Route Sweep: stopping here rather than iterating further

This check has now failed three times, each flagging a **different** pair of routes this branch does not touch:

| Run | Head | Flagged |
|---|---|---|
| 1 | `fa68f9b678` | `/platform/ai/overview`, `/platform/identity/agents`, `/platform/identity/authorization` |
| 2 | `88e2796bce` | `/admin/cockpit` 263→268, `/platform/edge-nodes` 158→162 |
| 3 | `53af5b8010` | `/platform/audit/authority` (aria shape), `/platform/tools/discovery` 963→970 |

**Run 1 was genuinely mine** and was fixed at source: I had written a ~60-word `capability_domain` paragraph into each new registry entry, and three roster pages render that field verbatim. Trimming it recovered 127 of the 219 added words. The irreducible remainder — three more coworkers really is three more rows — was then baselined from a measured `workflow_dispatch` run.

**Runs 2 and 3 are not mine, and no two runs agree on which routes they are.** The measurement is data-dependent, so unrelated drift reads as a regression on whichever branch is measured next. Run 2 is the clearest illustration: the dispatch refresh recorded `/admin/cockpit` at 263 while the PR run measured 268. Both numbers are real; committing the first tightened a floor the second immediately breached.

That is not convergeable by iteration, so I have stopped rather than push a fourth attempt. Filed as **BI-48741E66** (diff-scoped adjudication + drift tolerance; separate the ratchet from the gate).

### What this branch actually moved

```
/platform/identity/agents         4196 → 4340
/platform/ai/overview             3022 → 3114
/platform/identity/authorization  2453 → 2468
/workforce                         149 →  150
```

Seven changed lines: those four counts plus three `ariaSnapshot`s that grew because the accessibility tree gained roster rows. Nothing else differs from the pre-existing baseline — deliberately, because a branch should only move the baselines it causes to move.

### Everything else is green

**35 pass / 0 unrelated failures.** Unit tests (all four shards), Coworker Personas Audit, Prompt State-Leakage Audit, profession coverage, Policy Guards (source/workspace/PR), Typecheck, Production Build, DCO, CodeQL, seed-fit, UX-Fit.

CI caught six real defects in this branch along the way — missing personas, a prompt-leakage citation, a broken component test, an unmapped profession family, roster bloat, and a stale doc index — every one a genuine obligation of adding three registry identities. The gates worked. This last one is measuring something the branch cannot control.

Needs a human call: merge on the strength of the other 35 with BI-48741E66 tracking the gate, or hold until that lands.

markdbodman:
## Correction: I was wrong that this was unconvergeable

My previous comment claimed the UX Route Sweep was flagging a moving target of unrelated routes. That was wrong, and I should have checked before asserting it. The sweep **passes on four other open PRs** (#4758, #4757, #4749, #4748) and fails only here — so it is measuring something real about this branch.

**The miss:** when I scoped the measured refresh, I filtered on `defaultVisibleWords` only. That is why `/platform/audit/authority` never appeared in my list — its word count is **unchanged at 258**, but its **accessibility tree changed shape**, because the three new registry agents add authority rows. The sweep checks aria structure as well as word counts. A word-count filter is structurally blind to it, which is why run 3 looked like noise to me and wasn't.

Now baselined from the same measured dispatch run as the other four. Five routes qualify: four by word count, one by aria shape.

```
/platform/identity/agents         4196 → 4340   words
/platform/ai/overview             3022 → 3114   words
/platform/identity/authorization  2453 → 2468   words
/workforce                         149 →  150   words
/platform/audit/authority          258 →  258   aria tree reshaped
```

**Still deliberately not baselined:** `/platform/tools/discovery` (963 → 970 in one run). That route measured *lower* than its baseline in my dispatch sweep, so it is among the 17 this branch does not grow. Recording a number this branch cannot reproduce would be ratcheting on noise — the mistake I made two commits ago. If it flags again, that single route is drift, and the evidence for that is now on the record.

**The rule this branch follows:** take a measured value only where it *exceeds* the existing baseline, never where it falls below it.

BI-48741E66 stands, but its framing needs narrowing — the gate is not non-deterministic in the way I described. What is true is that a branch touching shared registry data affects more surfaces than are obvious, and the failure output makes aria-shape regressions easy to misread as word-count noise. I will update that item.

PR #4817, opened 2026-08-28
fix(guards): re-record unknown blocked_* statuses instead of dropping evidence (BI-BC57E1AC)

markdbodman:
## Summary

The local-CI client ships `blocked_*` statuses ahead of the deployed portal. `record_local_integration_result` rejected `blocked_child_signal_death` with `invalid_status`, and the run wrote **no evidence at all** — 25,553 passing tests vanished because the portal had not learned the status yet.

The only prior compatibility fallback was `blocked_sandbox_drift`. Any `blocked_*` rejected as `invalid_status` now re-records as `failed` with an infrastructure marker so the two load-bearing facts survive: the run produced no product evidence, and the reason was infrastructure.

Does not expand the portal enum (BI-C59AC8AF / BI-E2366B3B) and does not address host OOM.

## Verification

```
node --test scripts/lib/local-ci-unknown-blocked-status.test.mjs
```

6/6. Module-size ratchet unchanged.

Typecheck / production build / UX: not applicable (gate client helper).

## Docs

Docs-Impact-Decision: gate client only; no user-facing route or user-guide page.

Process-Spine-Decision: no new capability; the existing evidence write now degrades to a marked failed record instead of dropping the row.

## Local CI

Local-CI-Override: external-contribution-no-install: scripts-only node builtin guard; node --test 6/6; local-CI oversubscribed

Refs: BI-BC57E1AC

### DIFF
--- a/scripts/gate-worktree.mjs
+++ b/scripts/gate-worktree.mjs
@@ -59,6 +59,10 @@ import {
   isVerboseGateConsole,
 } from "./lib/pregate-console.mjs";
 import { isEntryModule } from "./lib/entry-module.mjs";
+import {
+  fallbackUnknownBlockedEvidence,
+  shouldFallbackUnknownBlockedStatus,
+} from "./lib/local-ci-unknown-blocked-status.mjs";
 
 const THIS_FILE = fileURLToPath(import.meta.url);
 const SCRIPT_DIR = dirname(THIS_FILE);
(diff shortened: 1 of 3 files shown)


markdbodman:
Closing as superseded by BI-C59AC8AF, already on `main`.

Main's fix covers this defect and more:

- It added `blocked_child_signal_death` to the canonical status set in `scripts/lib/local-integration-status.mjs` — the exact case this PR was opened for — so the status is no longer unknown to the recorder at all.
- It collapsed the four duplicate copies of that set (producer, MCP tool schema enum, handler runtime validation, recorder TypeScript union) into one module. Those four copies drifting apart is what caused the original defect.
- Its `fallbackStatusForUnknown` fires on **any** status the portal rejects as `invalid_status`. This PR's `shouldFallbackUnknownBlockedStatus` only fires when the status starts with `blocked_`, so main's version is strictly broader.
- `sandbox-freshness.test.mjs` on main already covers the fallback.

This PR's own module comment anticipated it: "Expanding the portal enum is a separate item (BI-C59AC8AF / BI-E2366B3B)." That item landed and absorbed the whole fix.

Merging this now would narrow the rule back to `blocked_*` and give it a second home alongside main's, so there is nothing left to salvage.

For the record, the two CI failures were real: `scripts/lib/local-ci-unknown-blocked-status.test.mjs` was never registered in `ci-policy-guards.mjs` (so CI would never have run it), and the branch was 49 commits behind main with a conflict in `gate-worktree.mjs` — the conflict being main's replacement of this very code.

PR #4844, opened 2026-08-29
fix(self-upgrade): survive a transient queue dispatch failure

markdbodman:
## What

An operator clicked "Upgrade now" at 2026-08-29 02:52 UTC. The run was created and immediately marked `failed`:

```
runId       | SUR-D71E8971
status      | failed
failureLog  | queue-dispatch-failed: fetch failed
startedAt   | (empty)
completedAt | 2026-08-29 02:52:29.67
```

The cause, from the portal log:

```
[TypeError: fetch failed]
  [cause]: Error [ConnectTimeoutError]: Connect Timeout Error
           (attempted address: inngest:8288, timeout: 10000ms)
    code: 'UND_ERR_CONNECT_TIMEOUT'
```

## Evidence that this was transient

- `dpf-inngest-1` had been up since 2026-08-27 13:33 and logged 4–9 events per minute continuously across 02:52, with no gap and no restart.
- Across the portal container's whole 2h20m uptime, `UND_ERR_CONNECT_TIMEOUT` appears **exactly once**, and `inngest:8288` is the only address it names.
- `fetch("http://inngest:8288/health")` from inside the portal returns 200 on demand.
- The identical dispatch succeeded unchanged on the next attempt (SUR-946F62CC).

One connect attempt lost a race against a 10-second timeout. There was no retry, no backoff, and no reaper for a run that died before `startedAt` was ever set — so it cost the operator the deploy and left a `failed` run that reads as though the upgrade itself broke. The install was at the time running a three-day-old image from a separate downgrade (BI-6CB35411), so the lost dispatch extended a live regression by over an hour.

## Why the runner's `retries: 0` does not settle it

`selfUpgradeManual` sets `retries: 0` deliberately, and this PR does not touch it: a *promotion* re-entered halfway is worse than one that failed.

That reasoning does not extend to the *enqueue*. Publishing an event is idempotent from the operator's point of view and has not yet touched the install — no quiescence drain, no image pull, no container recreate. The two steps have different safety properties and should not share one retry policy. Inngest, Temporal, Sidekiq/BullMQ and Step Functions all retry the publish at the client and keep work-level retry separate; DPF was stricter than any of them without gaining anything.

## Changes

- **`apps/web/lib/self-upgrade/dispatch.ts`** (new) — one shared helper for both call sites rather than two copies of a retry loop.
  - Retries only transport failures, walking the `cause` chain because undici throws a bare `TypeError("fetch failed")` with the real code nested underneath.
  - **Never** retries an HTTP status from the event API: a bad event key or malformed payload is a real misconfiguration and must surface immediately, not three attempts later.
  - Three attempts, ~1.25s total. The operator is watching a button.
  - `DispatchFailure` carries the attempts actually made, so the message cannot claim three tries when a non-retryable error stopped it at one.
- **`request.ts`** and **`promotions.ts`** — both dispatch sites wrapped; failure text now names the endpoint and error class instead of a bare `fetch failed`, so run history answers "is inngest up, or is the network wedged?" without a log dive.

## Verification

```
pnpm --filter web exec vitest run \
  lib/self-upgrade/dispatch.test.ts \
  lib/self-upgrade/request.test.ts \
  lib/actions/self-upgrade.test.ts

  Test Files  3 passed (3)
       Tests  32 passed (32)

pnpm --filter web typecheck    # clean
```

New coverage includes the exact nested undici shape that failed SUR-D71E8971, a first-attempt failure that then succeeds, three exhausted attempts, a 4xx that is not retried, and that backoff happens rather than spinning. `request.test.ts` gains an end-to-end case proving a transient timeout now queues instead of failing the run.

One existing assertion changed deliberately: it pinned the exact string `queue-dispatch-failed: inngest offline`. The message is now more informative, and the test asserts the underlying cause plus `after 1 attempt` — which also guards the bug I introduced and caught mid-review, where a fixed max-attempts constant made a single non-retryable failure read as three exhausted tries.

## Change classification

- [x] **Source-only change** (isolated unit logic; no route, schema, or UX surface)
- [x] Unit tests for affected files
- [x] Typecheck clean

Local-CI-Override: operator-emergency: operator directed skipping local gates for this change — the Windows pre-push gate does not run on this host and the local-CI pool is structurally single-slot. Unit tests and typecheck were run in the worktree; protected GitHub checks and the merge queue still apply.

## Governance note

The operator explicitly waived the initiative-readiness research and plan receipts for this item after four review dispatches to AGT-WS-PORTFOLIO failed to produce one. Root causes are recorded on BI-F0715C9C: the immutable source reader had no resolvable repository (`PROJECT_ROOT=/sandbox-workspace` is a synthetic single-commit bootstrap repo), and separately the reviewer does not emit its terminal writer call even after a successful read.

The design document is on the branch at `docs/superpowers/specs/2026-08-29-self-upgrade-dispatch-transient-resilience-design.md`.

Refs: BI-965E65B7

### DIFF
--- a/apps/web/lib/actions/promotions.ts
+++ b/apps/web/lib/actions/promotions.ts
@@ -52,6 +52,7 @@ import { inngest } from "@/lib/queue/inngest-client";
 import { readBuildPipelineLimit } from "@/lib/queue/admission";
 import { buildAdmissionSnapshot } from "@/lib/queue/admission-observability";
 import { SELF_UPGRADE_EVENT } from "@/lib/queue/functions/self-upgrade";
+import { describeDispatchFailure, inngestEndpoint, sendWithTransientRetry } from "@/lib/self-upgrade/dispatch";
 import { getErrorMessage } from "@/lib/shared/get-error-message";
 
 async function requireOpsAccess(): Promise<string> {
@@ -815,7 +816,7 @@ export async function triggerSelfUpgrade(opts?: { dryRun?: boolean; force?: bool
   });
 
   try {
-    await inngest.send({
+    await sendWithTransientRetry(() => inngest.send({
       name: SELF_UPGRADE_EVENT,
       data: {
         runId: run.runId,
@@ -823,10 +824,9 @@ export async function triggerSelfUpgrade(opts?: { dryRun?: boolean; force?: bool
         ...(opts?.dryRun !== undefined ? { dryRun: opts.dryRun } : {}),
         ...(opts?.force ? { force: true } : {}),
       },
-    });
+    }));
   } catch (err) {
-    const message = getErrorMessage(err);
-    await failRun(run.runId, `queue-dispatch-failed: ${message}`);
+    await failRun(run.runId, `queue-dispatch-failed: ${describeDispatchFailure(err, inngestEndpoint())}`);
     return { queued: false, reason: "queue-dispatch-failed", runId: run.runId } as const;
   }
 
--- a/apps/web/lib/self-upgrade/dispatch.test.ts
+++ b/apps/web/lib/self-upgrade/dispatch.test.ts
@@ -1,0 +1,115 @@
+import { describe, expect, it, vi } from "vitest";
+import {
+  DispatchFailure,
+  describeDispatchFailure,
+  isRetryableDispatchError,
+  sendWithTransientRetry,
+} from "@/lib/self-upgrade/dispatch";
+
+/** The exact shape undici throws — a bare TypeError with the real code nested in `cause`. */
+function connectTimeout(): Error {
+  const err = new TypeError("fetch failed");
+  (err as { cause?: unknown }).cause = Object.assign(
+    new Error("Connect Timeout Error (attempted address: inngest:8288, timeout: 10000ms)"),
+    { code: "UND_ERR_CONNECT_TIMEOUT" },
+  );
+  return err;
+}
+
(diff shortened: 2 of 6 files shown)


markdbodman:
Closing as superseded.

While this was being implemented, [#4840](https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/pull/4840) — `fix(self-upgrade): persist admission before dispatch (BI-3FD07259)` — merged to main at 06:07 UTC and solves the same defect more thoroughly.

This PR retried `inngest.send` so a transient connect timeout would not fail the run. Main instead **removes the coupling**: `requestSelfUpgrade` now persists a durable admission first and returns `dispatchStatus: admission_pending` without waiting on dispatch at all. A failed send becomes `indeterminate` (`queue-dispatch-indeterminate`) rather than a terminal `failed`, with `dispatchAttemptCount`, a dispatch lease, and a sweeper that re-attempts. Its own test now asserts `inngestSend` is never called on the request path.

That is the better design. Merging this would re-couple dispatch to the request path that #4840 deliberately decoupled, so the retry helper here is not worth salvaging on top.

The originating incident stands and is recorded on BI-965E65B7: SUR-D71E8971 was marked permanently `failed` by one connect timeout to inngest:8288 while inngest was up and processing events throughout. #4840 is what fixes it.

The design note on this branch remains accurate about the failure and the evidence, and the branch is left pushed as the record.

PR #4955, opened 2026-09-02
doc(spec): install public reachability — one contract, pluggable providers (BI-E2E8BD27)

markdbodman:
## The gap

The portal is fully built to sit behind a public hostname, and there is no supported way to give it one. `PUBLIC_URL`, `enforceCanonicalHost`, `PUBLIC_URL_ALIASES`, the Caddy TLS overlay and the step-ca PKI overlay all exist — and every one of them assumes a name that already resolves. On a residential connection behind CGNAT there is nothing to point DNS at, so that whole stack is unreachable infrastructure.

Verified against `origin/main` @ `8e262733ff`: a grep for `cloudflared|tailscale|ngrok|tunnel` across `docker-compose*.yml` and `scripts/` returns no reachability packaging. Nothing has shipped.

## What this is

A design spec only — no code. It answers **Open Question 2** of `2026-06-12-archetype-aware-mobile-companion-remote-access-design.md`, which left Cloudflare/Tailscale as "documented deployment recipes until tool evaluation approves a hard integration". That spec's decisions are carried forward unchanged and explicitly not reopened:

- tunnels move packets, not authorization
- the Authority Core remains the only policy decision point
- DPF does not depend on one relay vendor

## Research

Benchmarked against **Home Assistant Cloud (Nabu Casa)** as the comparable consumer product — a cloud relay proxying to an instance at home, no port forwarding, no DDNS — and three candidate providers.

**The discriminator is the operator's own domain.** Tailscale Funnel serves only `*.ts.net`, so it cannot put an install on a domain the operator owns; it stays valid as a private-mesh mode but is wrong as the first provider. A VPS reverse proxy can, but hands a consumer install a host to run and patch. **Cloudflare Tunnel** is outbound-only, serves the operator's own domain, supports many hostnames plus wildcards, and offers per-hostname Access policies.

A **DPF-operated relay is rejected outright** — it would make us custodian of every customer's traffic, against the single-tenant premise.

Following this repo's convention (step-ca, authentik), a tool evaluation under `docs/security/tool-evaluations/` is a **prerequisite of implementation**, not of this design.

## Safety invariants

1. The portal never learns which provider is in use, and works unchanged with none.
2. Only `portal:3000` is routable — 1455, 3001, 3035, 5433, 6379 and 8288-8289 must not be, and the overlay should make that structurally hard rather than documented.
3. Edge authorization is additive, never a substitute.
4. Enabling reachability is a visible operator act, not a hidden default — precisely the failure mode BI-006A04FE documents, where a default-off env var silently disabled federation while the UI reported success.
5. Removing the overlay returns the install to LAN-only with no residue.

## Open questions left explicit

Malformed `PUBLIC_URL` behind a public name (`parseCanonicalHost` currently fails open, silently accepting every Host header); where reachability state surfaces to the operator; and whether the first provider ships disabled-but-visible or absent.

## Verification

Every cited anchor re-checked against `origin/main`: `docker-compose.yml:180`, the canonical-host passthrough branch, `proxy.ts:30`, and the TLS overlay's 443/8443 listeners. Docs gates green — spec-status frontmatter OK, doc links PASS, prose lint clean (no new issues), doc index fresh, private-identity scan clean.

Blocks nothing; unblocks BI-E2E8BD27 implementation.

### DIFF
--- a/docs/superpowers/specs/2026-09-02-install-public-reachability-design.md
+++ b/docs/superpowers/specs/2026-09-02-install-public-reachability-design.md
@@ -1,0 +1,215 @@
+---
+status: active
+---
+
+# Install Public Reachability — one contract, pluggable providers
+
+**Date:** 2026-09-02 · **Epic:** EP-1FABA22D
+**Backlog:** BI-E2E8BD27
+
+## Problem
+
+The portal is fully built to sit behind a public hostname, and there is no supported
+way to give it one. Every piece of the last mile exists except the mile.
+
+A consumer install runs on a residential connection, usually behind CGNAT. There is
+no routable address to point DNS at, so the shipped TLS and PKI overlays — which
+assume a name that already resolves — are unreachable infrastructure. The operator
+who owns a domain and wants their install answerable on it has no path that the
+product supports.
+
+## Current-state anchors
+
+Verified against `origin/main` @ `8e262733ff`, 2026-09-02. The public-hostname mode
+is real and already wired:
+
+- `PUBLIC_URL` is plumbed through compose (`docker-compose.yml:180`) and read at
+  runtime by `resolveAppBaseUrl()` (`apps/web/lib/app-url.ts`) for outbound email
+  links, webhook payloads, and federation enrollment.
+- `enforceCanonicalHost()` (`apps/web/lib/canonical-host.ts`, called first in
+  `apps/web/proxy.ts`) 301s a non-matching origin to the canonical URL with
+  `Clear-Site-Data: "storage"`. `PUBLIC_URL_ALIASES` keeps LAN access alive
+  alongside a public name — the Home Assistant `internal_url` pattern.
+- `docker-compose.tls.yml` runs a Caddy sidecar terminating HTTPS on 443 to
+  `portal:3000`, plus an mTLS-only Edge action listener on 8443.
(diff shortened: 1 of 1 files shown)


markdbodman:
**Provider implementation is deferred and externally owned.**

Operator decision, 2026-09-02: the Cloudflare Tunnel work — the tool evaluation under `docs/security/tool-evaluations/` and the connector implementation — will be done by people who hold the account. It is not in this thread's scope.

This does not change the design. The whole point of the contract in §Design 1 is that the provider is selected by configuration and the portal never learns which one is in use, so the vendor-neutral half can be reviewed, approved and built independently of whoever implements a provider behind it.

What that leaves reviewable here:
- the reachability **contract** (`PUBLIC_URL`, `PUBLIC_URL_ALIASES`, `MCP_ALLOWED_ORIGIN_HOSTS` — all already consumed by the portal)
- the five safety invariants, particularly *only `portal:3000` is routable* and *edge authorization is additive, never a substitute*
- the three open questions

The provider comparison stays as recorded research — it is what the gate's `RESEARCH_REQUIRED` asked for, and it should still inform whoever picks the work up. Treat 'Cloudflare Tunnel as first provider' as a recommendation carrying its reasons, not a commitment this thread is making on their behalf.

PR #4970, opened 2026-09-02
fix(worktree-gate): a build the workspace decided about is classified

markdbodman:
## Why

A fresh worktree could not run `pregate` at all. The readiness probe reported

```
source-only · ignored_builds_unclassified:puppeteer,unrs-resolver,protobufjs
```

guard-parity preflight refused, and `pre-push` refused the recorded override too — offering a second bypass. Every code push from a newly created worktree hit this, and the only way through was stacking two emergency overrides.

**Those three are not unclassified.** `pnpm-workspace.yaml` decides them explicitly:

```yaml
allowBuilds:
  protobufjs: false
  puppeteer: false
  unrs-resolver: false
```

The worktree honouring that policy was the one being refused.

pnpm files a package under *"Explicitly ignored package builds (via `pnpm.ignoredBuiltDependencies`)"* only when the decision took that particular form. An `allowBuilds` decision still prints under *"Automatically ignored builds during installation"*, which the parser read as undecided. The `@scarf/scarf` fix (`259b3fb70f`, BI-0E472AE0) was correct and simply did not reach this case, because `@scarf/scarf` sits in the explicit section.

The sharpest evidence that the refusal was wrong: its stated reason is *"guards cannot run and would env-skip"* — and with `DPF_SKIP_COMPILE_READY_GATE` set, all 62 guards ran clean in 125s. They could run the whole time.

## What changed

- `classifyIgnoredBuilds(stdout, decidedPackages)` excludes packages the workspace already decided about, matching on the bare package name so the version and peer suffix (`puppeteer@25.7.0(yauzl@2.10.0)`) does not defeat the lookup.
- `readAllowBuildsDecisions()` reads the `allowBuilds:` block. Both `true` and `false` count as decided — a `true` package is built and never appears as ignored, a `false` one is ignored deliberately.
- `probeWorktreeReadiness` passes the decisions in.

**Fail-safe, deliberately.** An absent, malformed or unreadable block decides nothing, so the gate stays exactly as strict as before for anything genuinely undecided. A package nobody has decided about still fails closed — there is a test for that.

## Verification

Proven against the two live worktrees that were refused:

| | before | after |
|---|---|---|
| `D:\DPF-worktrees\animals-in-care` | `source-only` · 3 unclassified | `compile-ready` · `gateOk: true` · `ignoredBuilds: []` |
| this worktree | `source-only` · 3 unclassified | `compile-ready` · `gateOk: true` · `ignoredBuilds: []` |

- 32 tests pass in `scripts/lib/bootstrap-worktree-deps.test.mjs`, including a decided-package case, an undecided-package case that still fails closed, scoped-name and peer-suffix matching, and both fail-safe paths
- `pregate:preflight` — 62 guards clean
- **`pregate` — PASS, gated SHA `0f0d039a9670`, evidence `cmtjitbrd03rr01ofzmk37mh2`.** This branch was pushed with no override, which is itself the regression test: the gate that could not run now runs.

## Docs

`build-gate-runbook.md` already documents a previous failure of this exact class (the auto-heal that no-op'd on a partially-provisioned tree), so this one is written up beside it rather than attested away. `mcp-tool-authorization-runbook.md` links this file only for the bootstrap step and gains no user-visible change — attested in the commit trailer.

Closes BI-27DECD71. Unblocks BI-E54F7F87.


markdbodman:
Closing as redundant — and my diagnosis was wrong.

This was already fixed on `main` by #4965 (BI-4D0FCF3A), merged at 02:52 today while this branch was being written. That change classified the three packages into `pnpm.ignoredBuiltDependencies` in `package.json`, which is the mechanism pnpm's own reporting expects, and it resolves the problem completely: a worktree rebased onto current main probes `compile-ready` with `ignoredBuilds: []` and `gateOk: true`, with none of this branch's code.

My premise was also incorrect. I read the `allowBuilds:` block in `pnpm-workspace.yaml` and concluded the packages were already decided there and the classifier was blind to that mechanism. The repo's operative list is the `package.json` one; #4965 is the correct place for the decision. Teaching `classifyIgnoredBuilds` to read `allowBuilds` as well would have introduced a second source of truth for build decisions — a parallel utility for something that already has one home.

The one thing worth keeping from this branch is the observation, not the code: when the readiness probe refuses with "guards cannot run and would env-skip", that claim is falsifiable — with `DPF_SKIP_COMPILE_READY_GATE` set, all 62 guards ran clean in 125s. If that refusal wording is ever revisited, it overstates what is actually broken.

Retiring BI-27DECD71 as a duplicate of BI-4D0FCF3A.

PR #5141, opened 2026-09-06
feat(work-management): room definitions carry trigger, grant ceiling and measures; leaf profiles compose with the backbone (BI-EFFD97B4, BI-4B11F98E, BI-224E6E82, BI-58A9FC76)

markdbodman:
## What this delivers

Three pieces of the Workroom proactivity spine that sat unpushed on `feat/workroom-proactivity-spine` since 2026-08-28, rebased onto today's main:

- **BI-EFFD97B4** — every Workroom definition in the source registry carries a trigger (event / cadence / threshold, or null for imperative rooms), a tool-grant ceiling that can only narrow an agent's standing grants, and at least one measure with a binding key. `narrowRoomToolGrant` is the single intersection point.
- **BI-4B11F98E** — a leaf archetype value-stream profile composes with the universal backbone instead of replacing it (pet-rescue now projects 16 leaf + 5 backbone stages).
- **BI-224E6E82 / BI-58A9FC76** — archetype stages project into rooms and their measures resolve honestly through a metric-binding resolver (unresolvable binding reports unmeasurable, never zero).

Rebase follow-ups in this PR: the six registry entries added on main since the branch (coworker-engagement and the five employment-lifecycle definitions) now carry the same three fields; the registry module crossed the 800-line ceiling, so the vocabulary moved to `room-definition-contract.ts` and the registry-backed resolvers to `room-definition-resolvers.ts` (the registry re-exports the contract, so existing imports keep working); the stage-flow test expectation moved from 16 to 21 stages for the composed profile.

## Design grounding

Specs/plans reviewed: `docs/architecture/workroom-vocabulary-boundary.md` (deepen the registry, do not build a parallel template subsystem), the work-shape taxonomy spec and plan (#5009/#5059), and the room-definition identity work already on main. Code substrate reviewed: `apps/web/lib/work-management/source-registry.ts`, `room-types.ts`, `shape-projection.ts`, `apps/web/lib/twin/stage-flow.ts`. Source of truth: the source registry stays the single definition list; the new contract module holds only vocabulary and the pure intersection helper. Decision: extend the registry entries in place rather than introduce a second definition store.

## Verification

- `vitest run lib/work-management`: 64 files, 479 tests passed. `vitest run lib/twin`: 31 files, 185 tests passed.
- `pnpm run pregate:preflight`: OK, 64 guards clean (module-size ratchet satisfied after the split).
- `pnpm run pregate`: PASS for HEAD 269dc752df18 (slot-0). A first run on the pre-fix HEAD failed only on the stage-flow length expectation, fixed in the last commit.

Convergence-Impact-Decision: auto-converges — pure application code inside the portal image; the promoter's compose-from-release carries it to existing installs on the next self-upgrade, no data migration and no operator step.

Part of the 2026-09-06 worktree sweep.

The seeded content touched is the archetype value-stream composition in packages/storefront-templates: the rule that a leaf profile extends the universal backbone applies to every archetype, not one vertical; pet-rescue is only the profile the tests exercise.


markdbodman:
CI diagnosis for the repeated UX failure: the sole blocking route is `/admin/twin-kit`. Its frozen baseline is 479 words and this head renders 508, so the sweep classifies a confirmed regression (29 words) and blocks the budget/merge-readiness fan-in. All policy, source, workspace, typecheck, unit, ADP integration, and production-build gates pass. Because this PR changes the generated process projection and stage/room-definition content consumed by Twin Kit, treat the route delta as diff-related until disproven; reduce/defer the new above-the-fold wording or deliberately update the UX baseline only with matching reviewed UX evidence. Run: https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/34054815166/job/101545982048

markdbodman:
Follow-up on replacement head `0c017b2a`: the word regression is resolved (508 → 451, now only a 1-word advisory over the 450 budget), but the route remains blocked because `/admin/twin-kit` changed its frozen accessibility-tree shape by adding `group → button + text`. CI identifies this as a reviewed-baseline decision, not a flaky run. If the new disclosure is intentional, run the documented `ux-route-sweep.yml` baseline workflow and splice only `/admin/twin-kit` into the committed baseline; otherwise preserve the prior landmark/heading shape while deferring the detail. Job: https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/34057174033/job/101551125626

markdbodman:
All 37 GitHub checks now pass and review threads are clear, but `pnpm pr:health 5141` correctly still reports NOT READY: the latest head `a83c8194618f` has no local-CI sandbox evidence. The PR body references pregate evidence for the earlier `269dc752df18` head, which cannot be reused after the baseline commit. Please run the governed `pnpm run pregate` from the branch worktree and record the resulting evidence for `a83c8194618f`; do not use a docs-adjacent override because this PR changes application/runtime behavior.

PR #5168, opened 2026-09-07
fix(work): a completed stage earns the receipt that advances the shape (BI-76B35820)

markdbodman:
## Rooms dispatch forever and never advance

`nextStageKey` advances only when the current stage holds a receipt. Receipts were **read** in three places — the drive state reader, the plan resolver, the conformance observation — and **written in none**. Always `[]`, so every room re-dispatched its first stage on every tick, forever.

Measured on this install once the rooms started dispatching:

```
capsule       stage      completed runs
WC-A69BCABB   sweep      28
WC-6159CC0E   read       27
WC-6F4AE604   scan       27
WC-9B26C12E   classify   27
WC-C9ECBE09   read       27
WC-C9320161   assemble   25
WC-D40E9C3C   draft      25
WC-0A92C30D   read       24
WC-2845744F   sync       23
WC-12224280   measure    19
WC-21EBAD7B   read       15
WC-B02C8BFA   read       15
```

`COUNT(DISTINCT stageKey) = 1` on all twelve. `WC-A69BCABB` ran `sweep` 28 times and never reached `raise` or `decide`. `dependency-advisory-watch` exists to sweep advisories **and** raise findings **and** hand over the accept/patch/defer decision — it did step one, repeatedly, at a 15-minute cadence.

## Why it was hard to see

- **It looks healthy.** `action: dispatch_agent` on every room reads as success. Only the stage histogram shows the loop.
- **Real work happens.** The coworkers ran and completed. Nothing errored.
- **The stall source doesn't cover it.** `workroom-stall` watches `{pause, escalate}` — a room looping on stage one is *dispatching*, so it raises no attention. The estate burned model capacity every quarter hour while every surface stayed green.

## The rules

- **Only a completed run earns a receipt.** A failed or working run earns nothing, so the stage stays current and re-dispatches — the correct retry, and it stops a failure being laundered into progress.
- **Earning is idempotent.** 28 completed sweeps produce one receipt.
- **A cycle rollover clears receipts.** Otherwise a daily watch runs once and then satisfies every stage forever from yesterday's — the same defect inverted.
- **The correlation key has one builder.** `workroomStageTaskTitle` is used by the dispatcher that writes the task title *and* by the reader that looks for its completion. This is the eleventh read-with-no-writer in this program, and writer/reader drift is exactly how the class recurs — so the two callers share a function rather than a format.

## Verified

Title parse checked against the live database: it resolves all twelve rooms to their real completed runs (table above is that query's output).

**Known boundary, recorded in the code rather than gold-plated:** at the exact tick a cycle rolls over, resolution has already run against the previous cycle's receipts, so a room can advance one stage early. Bounded and self-correcting on the next tick; the alternative was deriving the cycle key in a second place.

## Design grounding

- Reviewed: `docs/superpowers/specs/2026-09-03-coordinated-workrooms-design.md` (source of truth), `work-shapes-and-the-decision-gate.md`, `drive-resolution.ts` (`nextStageKey`), `workroom-drive-state.ts`, `workroom-drive.ts`
- Decision: defect fix within the existing design; no new spec, no new contract. The receipt shape already existed — this gives it its first writer.

## Verification

- 65 pregate preflight guards clean
- 896 tests pass (104 files) — 13 new; two existing stored-state assertions updated for the added `lastCycleKey`
- `tsc --noEmit`: exit 0
- Live-database correlation verified

**The heavy local-CI build was NOT run and is NOT passed** — fenced on `host-capacity-lost:host-memory-low`. Recorded `operator-emergency` override; cloud merge queue is the adjudicating build per AGENTS.md §4.

`BI-76B35820`

Local-CI-Override: operator-emergency: heavy build unrun (NOT passed) — local-CI fenced on host-capacity-lost:host-memory-low. Fast tier clean (65 preflight guards, 896 tests, tsc --noEmit exit 0, live-database correlation). Cloud merge queue is the adjudicating build per AGENTS.md 4.

### DIFF
--- a/apps/web/lib/docs/doc-index.generated.json
+++ b/apps/web/lib/docs/doc-index.generated.json
@@ -2459,6 +2459,7 @@
         "who-answers-for-a-room",
         "nesting-has-to-be-written-not-just-declared",
         "who-may-coordinate-authority-and-a-gate-with-no-key",
+        "a-completed-stage-earns-a-receipt",
         "related-references"
       ]
     },
--- a/apps/web/lib/queue/functions/workroom-drive.ts
+++ b/apps/web/lib/queue/functions/workroom-drive.ts
@@ -16,6 +16,12 @@ import {
   resolveCoordinatorEligibility,
 } from "@/lib/work-management/coordinator-eligibility";
 import { planCoordinationBindings } from "@/lib/authority/coordination-bindings";
+import {
+  carryReceipts,
+  earnStageReceipts,
+  workroomStageTaskTitle,
+  type CompletedRun,
+} from "@/lib/work-management/stage-receipts";
 import { planContainmentRelations } from "@/lib/work-management/standing-room-nesting";
 
 import { gateAtEntry } from "../quiescence-gates";
@@ -55,6 +61,10 @@ export type WorkroomDriveRoom = {
   participants: ProjectableWorkroomParticipantAssignment[];
   currentStageKey: string | null;
   receipts: { stageKey: string; kind: string }[];
+  /** Task runs for this room's stages, used to earn stage receipts. */
+  stageRuns?: CompletedRun[];
+  /** The cycle the stored receipts belong to, so a rollover can clear them. */
+  lastCycleKey?: string | null;
   budgetUsage: { kind: string; used: number }[];
   stopConditionHits: string[];
   reviewDue: boolean;
@@ -139,7 +149,13 @@ export async function applyDrivePlan(input: {
     taskId: plan.taskId,
     lastRunAt: now.toISOString(),
     lastCycleKey: plan.cycle?.cycleKey ?? null,
-    receipts: room.receipts,
+    // A new cycle starts its stages again, so last cycle's receipts must not
+    // satisfy this cycle's — the same defect inverted.
+    receipts: carryReceipts({
+      receipts: room.receipts,
+      previousCycleKey: room.lastCycleKey ?? null,
+      currentCycleKey: plan.cycle?.cycleKey ?? null,
+    }),
     budgetUsage: room.budgetUsage,
     stopConditionHits: room.stopConditionHits,
     reviewDue: room.reviewDue,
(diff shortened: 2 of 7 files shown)


markdbodman:
## Do not merge this as the writer for BI-76B35820

PR #5166 already merged the fail-closed half of this defect (`2252c3eb4e` on `main`): after a `dispatch_agent` with no completing receipt, the next tick pauses with `executor_writeback_unavailable` and writes receipt kind `blocked` once. This branch is now `CONFLICTING` with that change.

The remaining gap is a **governed completing-receipt writer**. This PR is not that writer.

It treats `TaskRun.status === "completed"` as a completing receipt (`STAGE_COMPLETION_RECEIPT_KIND = "stage-run-completed"`). That was ruled out on the live install: coworkers claimed `Status: COMPLETE` with `executedTools=0` (WC-A69BCABB invented a clean sweep while `receipts: []`). A completed TaskRun is not outcome evidence. Earning a receipt from it would turn the fail-closed pause into silent false progress and keep burning capacity.

Keep this PR closed to merge until it is either:

1. Rewritten so TaskRun completion never earns a completing receipt, or
2. Superseded by a slice that lets the dispatched worker (or Process Overseer) record a completing receipt through MCP, while the drive still owns advance.

BI-76B35820 stays open until deploy of #5166 + live pause proof + independent PIR. The writeback slice is the next independently revertible PR, not this TaskRun-status earner.

markdbodman:
Agreed, and closing. The objection is correct and I verified it rather than taking it on trust.

**#5166 is on `main`** (`2252c3eb4`, with `workroom-drive-receipts.ts`), so the fail-closed half has landed.

**A completed TaskRun is not outcome evidence — confirmed with hard data on the live install:**

```
SELECT progressPayload->>'executedToolCount', count(*)
FROM "TaskRun" WHERE title LIKE 'Workroom WC-% / %' AND status='completed' GROUP BY 1;
 0 | 337
```

Every one of 337 "completed" runs executed **zero** tools. A representative payload:

```json
{"scheduledSummary": null, "executedToolCount": 0, "scheduledSummaryPayload": null}
```

So this PR would have converted 337 fabricated completions into stage advancement, and in doing so **undone the fail-closed pause #5166 just added** — trading a visible loop for silent false progress, which is strictly worse. The loop at least did not lie about outcomes.

My design error was treating the executor's self-reported status as evidence of the stage's outcome. That is provenance, not evidence, and the platform's own doctrine says so.

Closing rather than rewriting: option 2 in the review (a governed completing-receipt writer via MCP, with the drive still owning advance) is a different slice with a different shape, and it should be its own independently revertible PR rather than a rewrite of this branch.

`BI-76B35820` updated with the `executedToolCount: 0` evidence and this constraint. The zero-tool execution is a separate and larger defect than the advance gap — filing it accordingly.

PR #5206, opened 2026-09-08
fix(work-cases): a Workroom with no WorkItem is its own case, not a 404 (BI-2C31C399)

markdbodman:
You reported that opening a Workroom gives a 404. It does, for most of them.

**Measured on the live install: 464 Workrooms, 290 of them — 62% — carry no `workItemId`, and every one served the not-found boundary.**

## Why

The activity tree composed those links correctly. The failure was a layer down, and it is the kernel principle `a-room-is-reachable-by-construction` missing its second and third parts — the read model resolves a room through its anchoring foreign key, and a room that exists can be opened:

- `canonical-case-key.ts` needs an anchoring `workItemId` to redirect to, and returns null without one.
- `workspace-case-loader.ts` then looks for a WorkItem whose `sourceType` is `work-capsule`. Nothing is. `if (!item) return null` → `notFound()`.

A room with no WorkItem is not a broken anchor. It is a room that is its own unit of work, so it is its own case. An anchored room returns null from this path and still reaches its item's case, so one unit of work keeps one case (BI-EBEB77E2).

This follows the shape already in the loader, which branches to a dedicated projection for `coworker-engagement` — a second non-WorkItem source, not a new mechanism.

## What the case says, and what it won't invent

No recorded objective says so rather than restating the room's title as though that were intent. Urgency, effort and assignment read as "Not recorded" rather than taking a plausible default. The case is titled by the room's own title, because the shared summary would otherwise call it "Work capsule WC-ALPHA" — an identifier, not a name.

## Where it lives, and why not in the loader

`workspace-case-loader.ts` is at 798 of its 800-line module ceiling; adding the branch there pushed it to 818 and the ratchet refused, correctly. The logic lives in its own module and the route composes it, so the oversized module does not grow. The route is that loader's only caller.

## Verification

Against a seeded database, not only in unit tests: an unanchored room at `/workspace/cases/work-capsule%3AWC-NOITEM-1` now renders 2,281 characters carrying its real title, where before it served the not-found boundary. 10 new unit tests; the full work-management and workspace suites pass at 726. Typecheck and production build clean.

## Still broken, and recorded rather than claimed

An **anchored** room opened by capsule id still does not render — the other 174. `resolveCanonicalWorkCaseKey` returns the correct canonical key when called directly, and that canonical URL renders fine (2,129 characters), but the page's `redirect()` does not take effect: HTTP 200, no `Location` header, empty content. That is a distinct defect in the redirect rather than in this resolution path, and it is recorded on BI-2C31C399 with the evidence. This PR fixes the 62%, not the 38%.

Design-Grounding-Decision: implements the kernel principle a-room-is-reachable-by-construction (docs/founder-kernel/wiki/principles/a-room-is-reachable-by-construction.md, parts 2 and 3) at the resolution layer, following the non-WorkItem source pattern already in apps/web/lib/work-management/coworker-engagement-case-projection.ts; no new contract and no second case per unit of work.

Local-CI-Evidence: cmtrwri172uwc01qsnigxyxlo


markdbodman:
Correction to this PR's description.

I wrote that anchored rooms — the other 174 — "still do not render" and that the page's `redirect()` does not take effect. **That was wrong, and it was my measurement error, not a product defect.**

The redirect is issued after the shell has streamed, so Next delivers it as a client-side navigation rather than an HTTP 3xx. I probed with `waitUntil: "domcontentloaded"` and with `request.get(..., { maxRedirects: 0 })` — both read the response before that navigation completed, both saw a 586-character shell, and I read that as a broken page. Instrumenting the page proved `redirect()` was being called on every request all along.

Measured on current main, seeded database, real browser, waiting for the navigation to settle:

| Address | Result |
| --- | --- |
| `work-capsule:WC-NOITEM-2` (unanchored) | opens in place, 2,667 chars |
| `work-capsule:WC-UX-SWEEP` (anchored) | settles on `ux-sweep:ux-sweep-case` in ~500ms, 2,549 chars |

Both open. What this PR fixed — the 62% with no `workItemId` that genuinely served the not-found boundary — stands; that was verified by rendering, not by a probe racing a redirect.

Residual, recorded on BI-2C31C399 rather than claimed here: an anchored room paints an empty shell for about half a second, and a no-JavaScript request receives that shell rather than a 3xx. Resolving the canonical key in middleware would make it a true HTTP redirect. Soft issue, not a break.

PR #5213, opened 2026-09-08
fix(work): brief the stage, and advance it on evidence rather than a claim (BI-4A394B21, BI-76B35820)

markdbodman:
Two defects, one causal chain: the dispatched coworker had nothing to act on, and nothing it produced could advance the shape. Fixing either alone leaves the estate stuck, so they ship together.

## 1. The coworker was asked to do nothing (`BI-4A394B21`)

**337 dispatched stage runs reported `completed` with `executedToolCount: 0`** and no summary — every run, across all twelve standing rooms, with **zero evidence rows** between them.

The runs were real: quiescence run `QR-2026-09-07-e2d3ski9` was held by a live `coworker.reasoning-loop` for `Workroom WC-C9320161 / assemble`. Dispatch reached a coworker and a model ran. It had nothing to act on. This was the entire brief:

```
Execute Workroom WC-A69BCABB stage sweep for shape dependency-advisory-watch@1.0.0.
Stay inside the declared grants. Do not skip stages, widen authority, or invent occupants.
```

An opaque stage key and three prohibitions. Meanwhile the shape carried all of this, and none of it was sent:

| Shape declares | Was sent |
|---|---|
| stage title — "Sweep advisories against the manifest" | ✗ |
| `advance.condition` — *"Every advisory source in scope has been read and correlated to the recorded manifest"* (**the definition of done**) | ✗ |
| `evidence: ["assurance-run"]` — what to leave behind | ✗ |
| activity description, incl. "It never applies a patch" | ✗ |
| room objective, stop conditions | ✗ |

A model handed that will reasonably answer in prose that it did the work. 337 runs did.

The dispatcher now briefs from the shape, and says plainly that claiming completion advances nothing and that a blocker must be reported rather than faked.

## 2. Nothing the coworker produced could advance the stage (`BI-76B35820`)

A completed `TaskRun` is the executor's claim about **itself**. PR #5168 proposed earning the completing receipt from it and was correctly refused: it would have converted those 337 fabrications into stage advancement and undone the fail-closed pause from #5166. A visible loop is strictly better than silent false progress.

The receipt is earned instead from a **governed write the worker had to make through MCP** — `record_workroom_evidence`, a sanctioned mutator requiring `work_capsule_write` — carrying the stage it belongs to. The drive still owns the advance: a worker cannot advance itself, only leave evidence the drive reads.

Evidence must **name the stage**, be **of a kind the stage declared**, and **post-date the dispatch**. Anything short of that re-dispatches, visibly. There is no code path from run status to a receipt, and a test asserts that absence.

`record_workroom_evidence` gains an optional `stageKey`, with a schema/handler parity guard — the same seam already shipped broken once when `workShape` was advertised and silently dropped.

## Verified against the live database

The evidence query returns existing rows correctly, with `stageKey` null on all pre-existing evidence — which therefore earns nothing. Room-level notes are not stage outcomes.

## Module size

The substrate guard flagged `workroom-drive.ts` crossing the 800-LOC hotspot threshold as these loaders accumulated. Extracted to `workroom-drive-data.ts` — the drive decides, that module reads and writes what it decides over. **843 → 676 lines, and the hotspot set went 64 → 63.**

## Verification

- 66 pregate preflight guards clean
- **1608 tests pass** (178 files) — 23 new
- `tsc --noEmit`: exit 0
- Live-database query verification

**The heavy local-CI build was NOT run and is NOT passed** — fenced on `host-capacity-lost:host-memory-low`. Recorded `operator-emergency` override; cloud merge queue is the adjudicating build per AGENTS.md §4.

`BI-4A394B21` · `BI-76B35820`

Local-CI-Override: operator-emergency: heavy build unrun (NOT passed) — local-CI fenced on host-capacity-lost:host-memory-low. Fast tier clean (66 preflight guards, 1608 tests, tsc --noEmit exit 0, live-database verification). Cloud merge queue is the adjudicating build per AGENTS.md 4.

### DIFF
--- a/apps/web/lib/docs/doc-index.generated.json
+++ b/apps/web/lib/docs/doc-index.generated.json
@@ -2463,6 +2463,7 @@
         "nesting-has-to-be-written-not-just-declared",
         "who-may-coordinate-authority-and-a-gate-with-no-key",
         "concurrent-execution-and-recorded-state",
+        "the-brief-and-what-advances-a-stage",
         "related-references"
       ]
     },
--- a/apps/web/lib/mcp/packs/work-capsules-pack.ts
+++ b/apps/web/lib/mcp/packs/work-capsules-pack.ts
@@ -293,6 +293,13 @@ const definitions: ToolDefinition[] = [
       properties: {
         capsuleId: { type: "string", description: "Semantic Workroom id (WC-*)." },
         kind: { type: "string", enum: ENUMS.evidenceKinds, description: "Evidence kind." },
+        stageKey: {
+          type: "string",
+          description:
+            "The work-shape stage this evidence completes. REQUIRED for a Workroom stage to "
+            + "advance: the drive earns a completing receipt from stage-scoped evidence and "
+            + "never from a run's self-reported status (BI-76B35820).",
+        },
         summary: { type: "string", description: "Evidence summary." },
         command: { type: "string", description: "Optional command that produced the evidence." },
         url: { type: "string", description: "Optional URL for PRs, CI runs, screenshots, or external evidence." },
--- a/apps/web/lib/queue/functions/workroom-drive-data.ts
+++ b/apps/web/lib/queue/functions/workroom-drive-data.ts
@@ -1,0 +1,182 @@
+// Data loaders and reconcilers for the standing Workroom drive.
+//
+// Extracted from workroom-drive.ts, which the substrate-complexity guard
+// correctly flagged as crossing the 800-LOC hotspot threshold as these
+// accumulated. The drive decides; this module reads and writes what it decides
+// over.
+
+import { planCoordinationBindings } from "@/lib/authority/coordination-bindings";
+import {
+  COORDINATION_RESOURCE_TYPE,
+  COORDINATION_SCOPE_TYPE,
+} from "@/lib/work-management/coordinator-eligibility";
+import type { RecordedEvidence } from "@/lib/work-management/stage-evidence-receipts";
+import { planContainmentRelations } from "@/lib/work-management/standing-room-nesting";
+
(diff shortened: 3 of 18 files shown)


markdbodman:
Coordination from BI-CAP-AB63E190 / WC-DA928D2B: this overlaps the recovered fix/workroom-drive-writeback branch. I am holding publication of the separate receipt-writer mechanism to avoid shipping competing paths. Evidence-driven advancement is the better shared integration point, but I reproduced two blockers against f87d8b27ddde079e82d58f72c7a306c452409bd2:

1. earnEvidenceReceipts returns an existing blocked receipt unchanged even when fresh, matching stage evidence arrives. A pure function probe with existing [{stageKey:"read",kind:"blocked"}] and fresh findings returned only blocked. This leaves the currently paused estate stuck. Only an existing completing receipt should short-circuit; replace blocked when valid evidence earns completion.
2. stageHasCompletingEvidence accepts evidence with dispatchedAt:null. That can earn a generic receipt for a stage never dispatched, including human-attention stages. The generic evidence path must not substitute for the separately governed human approval path.

Also inspect the persistence seam: runWorkroomDriveJob passes earned receipts into resolveDrivePlan, but applyDrivePlan still receives the original room and builds its snapshot from room.receipts. Earned receipts therefore appear not to survive the tick. Please add a multi-tick regression, not only pure helper tests.

My branch includes first-failing regressions for human-attention receipt refusal, retaining task identity through a writeback pause, and preserving a completing receipt written between planning and persistence. Seventy focused tests and independent semantic review passed; full build is running. Please coordinate these protections into the single evidence-based path before merge. This is software correctness, not a request to weaken or add approval gates.

PR #5424, opened 2026-09-17
fix(ledger): the per-thread cost ledger reads a real number — thread attribution reaches telemetry and the CLI keeps its cache tokens (BI-CCF1ACBB)

markdbodman:
## What

`getThreadSpend` has returned zero for every coworker thread since it shipped in July (BI-CCF1ACBB, re-opened 2026-09-16). Measured on the reference install on 2026-09-17: 14,215 inference runs in 14 days, none carrying a `threadId`, and the ledger joins `AdapterRunTelemetry` on that column.

## Root causes (neither where the item body pointed)

1. **threadId was dropped upstream, not unwritten.** The telemetry writer and `callProvider` already accept and forward `attribution.threadId`. It never arrived: `routed-inference.ts` built the fallback-chain attribution without it, and `agentic-loop.ts` put `threadId` only on `mcpSession`, never on the route options. Both sites now carry it; `fallback.ts` also falls back to `mcpSession.threadId` the way it already did for `agentId`.
2. **"anthropic-sub reports 16 input tokens" was not a capture failure.** Those runs use the `claude-code-cli` adapter, and Anthropic's `input_tokens` is the *uncached* input only. The CLI parser discarded `cache_read_input_tokens` and `cache_creation_input_tokens`. It now keeps them; `ai-inference` already persists cache creation; the ledger folds uncached + cache-written + cache-read into `inputTokens` and exposes the split.

## Deliberately not done (recorded on the item)

- `ToolExecution.inputTokens` is not written: no tool result carries usage, and tools that infer internally already write their own telemetry rows, so writing tool tokens too would double count.
- `cachedInputTokens` persistence stays on BI-731F7FA2 (deliberate split). The runner reads the column and folds it as 0 until then.
- `estimatedCostUsd` is NULL on every row for every provider; pricing at read time is a follow-on.

## Closure

This item does **not** close on this merge. It closes when the liveness probe registered on BI-CCF1ACBB observes a real coworker thread with non-zero spend on a live install (BI-F6B8BADD: the probe observes, never gates).

## Governance

The implementation claim was refused `initiative_not_ready`: the readiness gate raised the shape small→large because the sensitivity heuristic matches the bare word "token" in an item about LLM tokens, and every recovery route is a reviewer coworker blocked by BI-EC82C48B; `declare_break_fix` was refused `break_fix_wip_exceeded`. Work proceeded under an operator override recorded on WC-F5A550FD. That gate is reported as **skipped under recorded authorization, never as passed**. DCO, pregate, PR protection and the merge queue are untouched.

## Verification

- Nine new unit tests fail on the base sources and pass with the fix (`thread-cost-ledger`, `cli-usage` + `cli-adapter`, `fallback`, `routed-inference`, `agentic-loop`). The new cases live in sibling test files and a small `cli-usage.ts` module so no baselined file grows and `cli-adapter.ts` stays under the 800-line ceiling.
- `pnpm --filter web typecheck` passes; pre-commit scoped typecheck passed; module-size and substrate-regression guards pass.
- Local-CI sandbox gate: **PASS** on 81290b02ff8e (evidence `cmu56cllx0fkq01prv3ghjoou`; typecheck, full vitest, production build compiled).
- Semantic review submitted once against the committed tree with a failure analysis citing that evidence: `TR-GATE-3683D111170E3C0C39CFB8F0` (policy is shadow mode; receipt recorded on the room when it lands).
- Workroom WC-F5A550FD carries the recorded operator override, the test evidence and the probe definition.

Docs-Impact-Decision: no user-facing route changed; the kernel principle's resolver note that described this machinery as inert is corrected in the same commit.
Convergence-Impact-Decision: auto-converges — promote.sh step 4b re-runs /docker-entrypoint.sh after every self-upgrade, whose seed (packages/db/src/seed.ts → seedWikiKernel, idempotent upsert of the founder-kernel wiki pages) carries the corrected principle page to every existing install; the code change ships in the image.

BI-CCF1ACBB


markdbodman:
Closing and reopening so CI re-reads the corrected PR-body trailer: a job re-run replays the original event payload (old body), which is why the re-run failed on text that no longer exists. Tracked as BI-E2669F37.

PR #5427, opened 2026-09-17
fix(self-upgrade): the N-1 health sentinel is a Compose service container, so the promoter replaces it instead of colliding with its name (BI-BC7AE37B)

markdbodman:
## Summary

The self-upgrade acceptance gate could not pass (BI-BC7AE37B). The N-1 baseline starts a health sentinel as `<project>-portal-1` with Compose project/service/number labels so it occupies the portal identity until the governed promoter replaces it. The promoter's `compose up -d --no-deps --force-recreate portal` (`scripts/promote.sh:857`) then failed with a name Conflict, and every step after "Resolve N-1 baseline" was skipped with no evidence artifact.

**Cause.** Compose selects a project's service containers by project + service + `oneoff=False`, and decides recreate-vs-keep from `config-hash`. The sentinel carried neither, so Compose filtered it out and tried to create a second container with the same name.

**Measured** on Docker 29.8.0 / Compose v5.5.1 with a minimal project and the promoter's exact `up` flags:

| sentinel labels | result |
|---|---|
| project + service + number | Conflict |
| + `oneoff=False` | Conflict |
| + `oneoff=False` + `config-hash` | adopted and recreated with the service image |

`version` is not required.

## Change

`sentinelComposeLabels(project)` adds `com.docker.compose.oneoff=False` and a `config-hash` Compose can never compute for the real service, so the candidate is always recreated and never adopted as current. The harness test asserts both labels; it failed on the unpatched harness (`sentinel must be a non-oneoff service container`) and passes now.

## Verification

- `node --test scripts/test-n-minus-one-upgrade.test.mjs scripts/check-n-minus-one-caller-honesty.test.mjs`: 35/35.
- `scripts/check-n-minus-one-caller-honesty.mjs`: OK.
- Acceptance workflow dispatched on this branch (`gh workflow run self-upgrade-acceptance.yml --ref fix/n-minus-one-container-name`); run link in the PR comments once it reports.
- Local-CI sandbox gate: **UNRUN** (PAT expired, lease claim returns 401).

Local-CI-Override: operator-emergency: install dpfmcp_ PAT expired, lease claim returns 401 (BI-78B653D5); pregate cannot run
Docs-Impact-Decision: CI harness internals, no user-facing page changes
Convergence-Impact-Decision: not-reachable — scripts/test-n-minus-one-upgrade.mjs is the CI acceptance harness and is never deployed to an install


markdbodman:
Acceptance evidence on d2a7ea28a: run 35248437305 (workflow_dispatch) resolved the N-1 baseline and ran the governed promotion end to end with no container-name Conflict — the defect this PR fixes. The job then died with exit 143 after nine silent minutes inside the promoter's cleanup sweep (image/builder prune); an earlier dispatch on the same SHA lost the hosted runner at 58 min in the same step. That is a different failure class and is filed as BI-DC04048A. https://github.com/OpenDigitalProductFactory/opendigitalproductfactory/actions/runs/35248437305

PR #5432, opened 2026-09-17
fix(self-upgrade): every cleanup sweep is bounded and named, and the harness stops a hung promoter with the last step in hand (BI-DC04048A)

markdbodman:
## Summary

With the N-1 sentinel fixed (#5427) the acceptance job ran the governed promotion end to end and then died inside the promoter's cleanup step: nine silent minutes after `step=cleanup`, then exit 143 with nothing to say which sweep hung or who stopped it (BI-DC04048A). An earlier dispatch on the same SHA lost the hosted runner at 58 minutes in the same step.

## Change

- **`scripts/promote.sh`** — the four cleanup sweeps (image prune, builder prune to the keep-storage cap, ephemeral image removal, superseded-tag removal) ran unbounded and unnamed. Each now runs under `_bounded_sweep`: its own `step=cleanup-<name>` line so the durable step trail names a hang, and a `timeout` (default 300 s, `PROMOTE_CLEANUP_SWEEP_TIMEOUT_SECONDS`) so a hang ends. A stopped sweep is reported and skipped; the upgrade stands, as the best-effort contract already promised. `step=cleanup-done` closes the section.
- **`scripts/test-n-minus-one-upgrade.mjs`** — parsed `--timeout-ms` (default 20 min) and never applied it. The promoter run now carries that bound, and `execFileWithLiveOutput` remembers the last `step=` line so a child stopped by any signal is reported as `stopped by SIGTERM after the harness's bound (last promoter step seen: cleanup)` instead of a bare 143.

No environment key is added to the promoter invocation, so the caller-honesty guard is unaffected.

## Verification

- `node --test` harness, caller-honesty and promoter-contract suites: 40/40. New test covers the bound reaching `execFile` and the last-step diagnosis.
- `check-n-minus-one-caller-honesty.mjs`: OK. `bash -n promote.sh`: OK.
- Acceptance workflow dispatched on this branch; result posted in a comment when it reports.
- Local-CI sandbox gate: **UNRUN** (PAT expired; no credentials client issued yet).

Local-CI-Override: operator-emergency: install dpfmcp_ PAT expired and no client_credentials client issued yet (BI-78B653D5); pregate cannot run
Docs-Impact-Decision: promoter internals and CI harness; no user-facing page changes
Convergence-Impact-Decision: auto-converges — promote.sh ships inside the candidate promoter image, so every install picks it up on its next self-upgrade


markdbodman:
Acceptance run on 18f9887ca (35270554727, workflow_dispatch): INCONCLUSIVE. The hosted runner lost communication with the server 59 minutes into the promotion step and GitHub retained no log, so neither the new per-sweep step trail nor the harness's 20-minute bound produced readable evidence. This is the same shape as run 35239126422 on #5427 and points at runner resource starvation during the promotion (the step builds the portal image with an 8 GB heap plus the sandbox image on a hosted runner), not at the cleanup sweep alone. The change here still stands on its own: unbounded sweeps are now bounded and named, and a hung promoter is reported with its last step. Getting this gate green likely needs a larger runner or a split of the image build out of the promoter step; that is an operator decision.

